PR #2, opened 2019-11-14
Refreshing of a target database using dumps from our production database

tsibley:
This is useful for refreshing our shared testing instance, but can also be used for local dev instances.  The process is intended to be robust so that we could put it on an automatic weekly or (even daily?) schedule if desired.

Command for refreshing your local dev environment is:

```
./dev/refresh-database
```

Command for refreshing the testing environment is:

```
PGHOST=testing.db.seattleflu.org PGDATABASE=testing PGUSER=postgres ./dev/refresh-database
```

If you provide a directory as the first argument, the database dump will be preserved and re-used between runs (so you can save a bunch of time and use the same dump for a week or two for local dev).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,18 @@
-# Backoffice interfaces
+# Backoffice
 
-Interfacing with and tying together the various backoffice services.
+Services:
+
+  * [ID3C production environment](id3c-production/README.md)
+  * [ID3C testing environment](id3c-testing/README.md)
+  * [Metabase](metabase/README.md)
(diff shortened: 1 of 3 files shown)


tsibley:
One note: If kicking the tires on this, please don't run it on our actual testing database right now. Limit to your dev database. We expect NWGC to be using testing today/tomorrow to send us some example data, and don't want to blow that away. :-)

tsibley:
Pushed some README changes. @kairstenfay Thanks for pushing me to do that. :-)

PR #4, opened 2019-11-26
Automate upload of new manifest data

kairstenfay:
Automate the upload of new manifest data by adding a new cronjob that invokes the `update-unattended` script from http://github.com/seattleflu/specimen-manifests.

### DIFF
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -36,6 +36,9 @@ ENVD=/opt/backoffice/id3c-production/env.d
 # conditions.
 35 * * * * ubuntu envdir $ENVD/slack/ pipenv run id3c reportable-conditions notify --commit
 
+# Upload new manifest data
+40 * * * * ubuntu envdir $ENVD/hutch/ /opt/specimen-manifest/update-unattended
+
 # Manifests are manually uploaded periodically right now, so just check once an
 # hour at quarter 'till.
 45 * * * * ubuntu pipenv run id3c etl manifest --commit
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Automate manifest ingestion from S3 bucket into ID3C](https://trello.com/c/WRqVCjrT/267-automate-manifest-ingestion-from-s3-bucket-into-id3c)

crontabs/id3c-production
```diff
@@ -32,10 +32,13 @@ ENVD=/opt/backoffice/id3c-production/env.d
...
 30 * * * * ubuntu pipenv run id3c etl presence-absence --commit
 
-# After presence/absence results are uploaded, check if any contain reportable 
-# conditions. 
+# After presence/absence results are uploaded, check if any contain reportable
+# conditions.
 35 * * * * ubuntu envdir $ENVD/slack/ pipenv run id3c reportable-conditions notify --commit
 
+# Upload new manifest data
+40 * * * * ubuntu /opt/id3c-production/specimen-manifest/update
```
tsibley: I think better to make this `/opt/specimen-manifest` since the manifest repo is not tied to either production or testing infra. Or maybe better still would be your first suggestion of a clone made every time by the `update-unattended` script?

kairstenfay: We have discussed this already, but the latest decision is to keep the update script within the `specimen-manifests` repo. Therefore, we pull every time within the update-unattended script rather than cloning.

crontabs/id3c-production
```diff
@@ -36,6 +36,9 @@ ENVD=/opt/backoffice/id3c-production/env.d
 # conditions.
 35 * * * * ubuntu envdir $ENVD/slack/ pipenv run id3c reportable-conditions notify --commit
 
+# Upload new manifest data
+40 * * * * ubuntu /opt/specimen-manifest/update-unattended
```
kairstenfay: Does this need to be wrapped with "pipenv run"?

PR #6, opened 2019-12-03
Stop relying on default AWS credentials and instead use environment vars

tsibley:
This now parallels the way our other set of AWS credentials (for Fred
Hutch) works.  It will also make it similar to how we'll use
EC2-instance-provided credentials in the future.


tsibley:
Deployed.

PR #7, opened 2019-12-16
science-map: Update Y2 start date to Oct 01

kairstenfay:
We had previously set a Year 2 start date arbitrarily to September 1,
2019. Per Trevor, we want to update this start date to October 1, 2019
to "make our use of 'Y1' synonymous with the use of '2018-2019 influenza
season' by outside groups."


trvrb:
Great! Again, I think the best thing here will be a date selection slider with min and max like what auspice has and default to something like min of -30 days and max of present. Not a big hurry.

kairstenfay:
> Great! Again, I think the best thing here will be a date selection slider with min and max like what auspice has and default to something like min of -30 days and max of present. Not a big hurry.

This task exists as a Trello card in our backlog: https://trello.com/c/pa8EuS9g.

PR #12, opened 2020-01-30
Start Metabase and Lab Labels services at boot

tsibley:
This was the intended behaviour, but it was thwarted by using an
incorrect, non-existent target unit name (local.target).  Whoops!

Use a standard built-in target instead!

Automatically starting services after boot is important since the
backoffice will now, if necessary, automatically reboot in the middle of
the night to apply security patches.

### DIFF
--- a/lab-labels/lab-labels.service
+++ b/lab-labels/lab-labels.service
@@ -9,4 +9,4 @@ ExecStart=/usr/bin/docker start -a lab-labels
 ExecStop=/usr/bin/docker stop -t 2 lab-labels
 
 [Install]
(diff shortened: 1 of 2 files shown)


tsibley:
Deployed with:

```
cd /opt/backoffice
git pull
sudo systemctl daemon-reload   # Reload symlinked {metabase,lab-labels}.service unit files
sudo systemctl enable metabase
sudo systemctl enable lab-labels
systemctl list-dependencies default.target  # Verify {metabase,lab-labels}.service appear
```

PR #13, opened 2020-02-04
science-map: Update to augur build metadata v3

kairstenfay:
Update the export script for the science map to use the new version of
the augur build metadata view.

~This PR is dependent on https://github.com/seattleflu/id3c-customizations/pull/40 and will be kept as a draft until the original PR is merged.~

### DIFF
--- a/export/science-map-geojson
+++ b/export/science-map-geojson
@@ -15,10 +15,10 @@ psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
         )
     )
 
-    from shipping.metadata_for_augur_build_v2
+    from shipping.metadata_for_augur_build_v3
     join warehouse.tract on (tract.identifier = residence_census_tract)
 
     where date >= '2019-10-01'
       and age_category in ('child', 'adult')
-      and site_category in ('clinical', 'community')
+      and site_category in ('clinic', 'community', 'hospital')
 "


kairstenfay:
[Start using `metadata_for_augur_build_v3` for the science map export.](https://trello.com/c/49PTxpwm/333-start-using-metadataforaugurbuildv3-for-the-science-map-export)

tsibley:
> Also, because age and site categories are no longer ad-hoc, switch from
> hard-coding acceptable categories to filtering out ones that are null.

I'd hardcoded the categories because the science map itself hardcodes the filters. If we add a new age category, say `tween` for example, then there won't be a checkbox on seattleflu.org/science for `tween` and unchecking both adult and child will still show tweens. Is that behaviour an improvement over never showing tweens? Maybe!

kairstenfay:
> I'd hardcoded the categories because the science map itself hardcodes the filters. If we add a new age category, say `tween` for example, then there won't be a checkbox on seattleflu.org/science for `tween` and unchecking both adult and child will still show tweens. Is that behaviour an improvement over never showing tweens? Maybe!

Hmm I see. 
The upside of hardcoding the categories is that we guarantee which age or site categories enter the science map. 
The downside is that we have some (albeit minimal) extra code to maintain.

tsibley:
Nod. I think it's six of one, half a dozen of the other. Just wanted to note the original motivation for the explicit where clause.

kairstenfay:
> Nod. I think it's six of one, half a dozen of the other. Just wanted to note the original motivation for the explicit where clause.

It feels like a toss-up to me regarding which method is preferable. I'm happy to take the path of least resistance here, meaning, I am ok with making the fewest changes to the existing code as possible. 
Thus, I can update the site categories to the new ones found in the view and keep the hard-coding in place.

kairstenfay:
[Remove metadata_for_augur_build_v2](https://trello.com/c/i5khEuyi/340-remove-metadataforaugurbuildv2)

kairstenfay:
FYI I changed the psql export query back to checking for hard-coded age and site categories.

export/science-map-geojson
```diff
@@ -15,10 +15,10 @@ psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
...
     )
 
-    from shipping.metadata_for_augur_build_v2
+    from shipping.metadata_for_augur_build_v3
     join warehouse.tract on (tract.identifier = residence_census_tract)
 
     where date >= '2019-10-01'
       and age_category in ('child', 'adult')
-      and site_category in ('clinical', 'community')
+      and site_category site_category in ('clinic', 'community', 'hospital')
```
joverlee521: Rework typo?

kairstenfay: Yes, thank you.

PR #15, opened 2020-02-06
Add routine to terminate idle database sessions

kairstenfay:
Add a crontab that runs every minute to terminate idle database sessions
other than those of superusers, the `rdsadmin` and the `metabase` roles
that have been idle for more than 30 minutes.

### DIFF
--- a/bin/terminate-idle-sessions
+++ b/bin/terminate-idle-sessions
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -euo pipefail
+psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
+    with candidate_connections as (
+        select
+            pid,
+            client_addr,
+            usename,
+            application_name,
+            current_timestamp - state_change as idle_duration
+        from
+            pg_stat_activity_nonsuperuser
+        where
+                pid != pg_backend_pid()
+            and state in ('idle', 'idle in transaction', 'idle in transaction (aborted)', 'disabled')
+        order by
+            client_addr,
+            application_name,
+            state_change desc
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Idle disconnect for database](https://trello.com/c/ygWRstQJ/322-idle-disconnect-for-database)

export/terminate-idle-sessions
```diff
@@ -0,0 +1,27 @@
...
+            application_name,
+            state_change desc
+    )
+
+    select *, pg_terminate_backend(pid)
+    from candidate_connections
+    where idle_duration > interval '30 min'
+    and usename not in ('rdsadmin', 'metabase')
+    and usename not in (select rolname from pg_roles where rolsuper = 't')
+"
```
tsibley: This will end up alerting us on Slack whenever it does terminate an idle session, which may or may not be useful. Thoughts?

kairstenfay: Should we add the `chronic` command to silence it? Or will a notification still be generated?

tsibley: chronic only sends the output it buffers if the command exits with a non-zero exit status, so using it here would suppress all output unless there's an error. I don't expect there would be with the current query, even if `pg_terminate_backend` returns false, indicating it failed.

Another option, instead of using chronic, is to adjust the query to only report failed terminations. This might be nicer?

kairstenfay: I updated the query to return only results with `pg_terminate_backend(pid) = 'f'`. If the table is empty (no failed terminations), will cron still produce an email?

PR #16, opened 2020-02-11
Import new UW retrospectives into the REDCap project used to pull clinical data

tsibley:
Parses the standard BBI-managed manifests to extract the info we need,
diffs to identify new records, and then uploads those to REDCap via the
API.

Runs in a dry run mode by default which just prints the JSON which would
be imported.  Pass --import to actually do the import.

---

This is reviewable and testable now (be careful if using `--import`), but there a still a couple things left for me to do:

- [x] Add usage doc to the top of the program
- [x] Add a cronjob to `id3c-production`


tsibley:
Reviewable now.

bin/import-uw-retrospectives-to-redcap
```diff
@@ -0,0 +1,161 @@
...
+    if [[ $import == import-to-redcap ]]; then
+        import-to-redcap < <(<"$new_records" as-redcap-records)
+        cp-records "$latest_records" "$previous_records"
+    else
+        <"$new_records" as-redcap-records
+    fi
+}
+
+parse-manifest() {
+    id3c manifest parse-using-config "$CONFIG" \
```
joverlee521: When I try to run with just `bin/import-uw-retrospectives-to-redcap`, I get `line 95: id3c: command not found`.
Am I missing something?

tsibley: Ahh, should have documented this in the usage. The script expects to be run under `pipenv run`.

This is a bit of a change from our other scripts which handle `pipenv` themselves, and I think will be a nicer separation of concerns. (So may update those in time to do the same.)

joverlee521: Ahh, I see. It makes sense to pull `pipenv` out of the script itself.

tsibley:
Bulk imported into REDCap all the records so Becca can test her process out.

bin/import-uw-retrospectives-to-redcap
```diff
@@ -0,0 +1,110 @@
...
+    parse-manifest > "$latest_records"
+    diff-manifests "$previous_records" "$latest_records" > "$new_records"
+
+    if [[ ! -s $new_records ]]; then
+        echo "No new records" >&2
+        exit
+    fi
+
+    if [[ $import == import-to-redcap ]]; then
+        import-to-redcap < <(<"$new_records" as-redcap-records)
```
kairstenfay: What is happening on line 52?

tsibley: The shell feature `<(…)` forks off the command `…` in a separate process and provides a named pipe like `/dev/fd/42` in its place. `a < <(b)` sets the stdin of command `a` to the output of command `b`.

I liked the way this line read left-to-right as "import to redcap the new records as redcap records", but maybe the syntax is too unclear.

If it's helpful, these are all equivalent:

```bash
import-to-redcap < <(<"$new_records" as-redcap-records)

import-to-redcap < <(as-redcap-records <"$new_records")

as-redcap-records <"$new_records" | import-to-redcap

<"$new_records" as-redcap-records | import-to-redcap

cat "$new_records" | as-redcap-records | import-to-redcap
```

I can change this to something more typical if you want.

kairstenfay: > I can change this to something more typical if you want.

No, it's ok. I was just curious what the syntax was doing. Thanks for the explanation.

tsibley:
Merged and deployed, provisioned with my REDCap API token (since we can't have service role accounts).

PR #17, opened 2020-02-25
crontabs: Configure jobs for swab-and-home-flu

joverlee521:
Not entirely sure if the timing of the swab-n-send etls. I currently have it set so that the old project is processed at five past and the new project is processed the rest of the time. I'm hoping eventually we can just delete the old project job when it has completely stopped.


crontabs/id3c-production
```diff
@@ -64,10 +64,13 @@ ENVD=/opt/backoffice/id3c-production/env.d
...
 # Transform REDCap DETs to FHIR documents every 10 minutes, interleaving kiosk
-# and swab-n-send on the 5s (since they share a cache file).  Ingest FHIR
-# documents every 5 minutes, scheduled 3 minutes after each DET run.
+# and swab-n-send on the 5s (since they share a cache file). Only process the
+# old swab-n-send project at 5 past and process from new project the rest of time.
+# We can eventually delete the old project once it has been completely stopped.
+# Ingest FHIR documents every 5 minutes, scheduled 3 minutes after each DET run.
 0-50/10 * * * * ubuntu flock -F /home/ubuntu/cache.pickle envdir $ENVD/deidentification envdir $ENVD/smartystreets envdir $ENVD/redcap-kiosk pipenv run id3c etl redcap-det kiosk --commit
-5-55/10 * * * * ubuntu flock -F /home/ubuntu/cache.pickle envdir $ENVD/deidentification envdir $ENVD/smartystreets envdir $ENVD/redcap-swab-n-send pipenv run id3c etl redcap-det swab-n-send --commit
+5 * * * * ubuntu flock -F /home/ubuntu/cache.pickle envdir $ENVD/deidentification envdir $ENVD/smartystreets envdir $ENVD/redcap-swab-n-send pipenv run id3c etl redcap-det swab-n-send --commit
```
kairstenfay: I think we wouldn't want this old swab-n-send job to run at 5 past because the swab-n-send and kiosk projects run on the 5s.

joverlee521: I'm interpreting it as the two are interleaving on the 5s, so kiosks run at 0, 10, 20, 30, 40, 50 and the swab-n-send runs at 5, 15, 25, 35, 45, 55. 
Here I've made the old project run at 5 and the new project run at 15, 25...

kairstenfay: Oh I see that now! I find this syntax difficult to read, but it looks like you did the right thing.

joverlee521:
Closing in favor of #19

PR #20, opened 2020-03-03
Use our custom logging configs for our API servers running under uWSGI

tsibley:
Makes sure that syslog messages from them are handled the same way.


tsibley:
Whoops: https://github.com/seattleflu/backoffice/commit/4a9d6673439e4a66f2c06d36b3855c5bdc8bdf03

PR #23, opened 2020-03-12
Add routine to terminate old metabase sessions

joverlee521:
Add a crontab that runs every minute to terminate old metabase sessions that have been active for more than 10 minutes.

10 was an arbitrary number, we can change the interval if we need to.

### DIFF
--- a/bin/terminate-old-metabase-sessions
+++ b/bin/terminate-old-metabase-sessions
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -euo pipefail
+psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
+    with active_metabase_connections as (
+        select
+            pid,
+            current_timestamp - state_change as active_duration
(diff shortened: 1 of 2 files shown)


joverlee521:
[Kill active metabase queries open for too long](https://trello.com/c/xhCStw72/391-kill-active-metabase-queries-open-for-too-long)

joverlee521:
> This is fine, although I really would like to fix the underlying views sooner than later.
> 
> I do also wonder if this will mess up internal queries Metabase runs against the database. It does a bunch of background things like, for example, periodically scanning distinct column values for all tables/views in order to provide autocomplete. We should be cognizant of this possibility if we notice other issues with Metabase. Errors would show up in its logs, but those logs don't surface anywhere we routinely look.

From what I've seen, it seems like the internal Metabase queries become `idle`. This only terminates `active` queries that have not changed state for 10 minutes.

PR #25, opened 2020-03-20
Scan ror

kairstenfay:
Adds a workflow for exporting REDCap SCAN data joined with ID3C SCAN return of results data.

Depends on https://github.com/seattleflu/id3c-customizations/pull/71

Opening as a draft until we know the location of the output S3 file.

### DIFF
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -8,8 +8,8 @@ set -euo pipefail
 : "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
 : "${REDCAP_API_TOKEN:?The REDCAP_API_TOKEN environment variable is required.}"
 
-fields="participant_first_name,participant_last_name,birthday,return_utm_barcode,contacted"
-filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [return_utm_barcode] != ''"
+fields="participant_first_name,participant_last_name,birthday,pre_scan_barcode,utm_tube_barcode_2,reenter_barcode,return_utm_barcode,contacted"
+filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != ''"
 
 curl -fsSL -H "Content-Type: application/x-www-form-urlencoded" \
       -H "Accept: application/json" \
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -1,0 +1,31 @@
+#!/bin/bash
+# usage: export-redcap-scan
+#
+# Exports REDCap the given *fields* from all projects matching the given
+# *api_token*.
+#
+# Does not export field names (csv header).
+#
+set -euo pipefail
+
+: "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
+
+fields="${1:?A comma-separated string of requested fields from REDCap is required as the first argument.}"
+api_token="${2:?A REDCap API token is required as the second argument.}"
+
+filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [return_utm_barcode] != ''"
+
+# Retrieve specified REDCap data, removing csv header
+awk '{if (NR!=1) {print}}' < \
+<(curl -H "Content-Type: application/x-www-form-urlencoded" \
+      -H "Accept: application/json" \
+      -X POST \
+      -d token="$api_token" \
+      -d format=csv \
+      -d content=record \
+      -d type=flat \
+      -d fields="$fields" \
+      -d filterLogic="$filters" \
+      -d exportDataAccessGroups=false \
+      -d returnFormat=json \
+      "$REDCAP_API_URL")
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -11,7 +11,7 @@ set -euo pipefail
 fields="participant_first_name,participant_last_name,birthday,return_utm_barcode,contacted"
 filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [return_utm_barcode] != ''"
 
-curl -H "Content-Type: application/x-www-form-urlencoded" \
+curl -fsSL -H "Content-Type: application/x-www-form-urlencoded" \
       -H "Accept: application/json" \
       -X POST \
       -d @<(echo "token=$REDCAP_API_TOKEN") \
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -1,23 +1,17 @@
 #!/bin/bash
 # usage: export-redcap-scan
 #
-# Exports REDCap the given *fields* from all projects matching the given
-# *api_token*.
-#
-# Does not export field names (csv header).
+# Exports participant information from the SCAN REDCap projects as CSV.
 #
 set -euo pipefail
 
 : "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
 : "${REDCAP_API_TOKEN:?The REDCAP_API_TOKEN environment variable is required.}"
 
-fields="${1:?A comma-separated string of requested fields from REDCap is required as the first argument.}"
-
+fields="participant_first_name,participant_last_name,birthday,return_utm_barcode,contacted"
 filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [return_utm_barcode] != ''"
 
-# Retrieve specified REDCap data, removing csv header
-awk '{if (NR!=1) {print}}' < \
-<(curl -H "Content-Type: application/x-www-form-urlencoded" \
+curl -H "Content-Type: application/x-www-form-urlencoded" \
       -H "Accept: application/json" \
       -X POST \
       -d @<(echo "token=$REDCAP_API_TOKEN") \
@@ -28,4 +22,4 @@ awk '{if (NR!=1) {print}}' < \
       -d filterLogic="$filters" \
       -d exportDataAccessGroups=false \
       -d returnFormat=json \
-      "$REDCAP_API_URL")
+      "$REDCAP_API_URL"
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -9,9 +9,9 @@
 set -euo pipefail
 
 : "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
+: "${REDCAP_API_TOKEN:?The REDCAP_API_TOKEN environment variable is required.}"
 
 fields="${1:?A comma-separated string of requested fields from REDCap is required as the first argument.}"
-api_token="${2:?A REDCap API token is required as the second argument.}"
 
 filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [return_utm_barcode] != ''"
 
@@ -20,7 +20,7 @@ awk '{if (NR!=1) {print}}' < \
 <(curl -H "Content-Type: application/x-www-form-urlencoded" \
       -H "Accept: application/json" \
       -X POST \
-      -d token="$api_token" \
+      -d @<(echo "token=$REDCAP_API_TOKEN") \
       -d format=csv \
       -d content=record \
       -d type=flat \
--- a/bin/scan-return-of-results/transform
+++ b/bin/scan-return-of-results/transform
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+"""
+Join the REDCap csv data with an ID3C csv.
+Parses the joined data to make it suitable for UW Lab Med's return of results
+portal.
+"""
+import sys
+import argparse
+import pandas as pd
+from pathlib import Path
+
+
+def parse_redcap(redcap_file) -> pd.DataFrame:
+    """
+    Reads in data from a given *redcap_file*. Returns a pandas.DataFrame
+    prepared in the specifications of UW Lab Med's return of results portal.
+    """
+    redcap_data = pd.read_csv(redcap_file)
+    redcap_data = redcap_data.rename(columns={'return_utm_barcode': 'qrcode', 'birthday': 'birth_date'})
+
+    participant_name = lambda row: f"{row['participant_first_name']} {row['participant_last_name']}"
+    redcap_data['pat_name'] = redcap_data.apply(participant_name, axis='columns')
+    redcap_data = redcap_data.drop(columns=['participant_first_name', 'participant_last_name'])
+
+    return redcap_data
+
+def edit_status_code(scan_data: pd.DataFrame) -> pd.DataFrame:
+    """
+    Sets the *scan_data* status to `never-tested` in the unlikely event that a
+    REDCap record's barcode is not found in ID3C.
+
+    Sets the *scan_data* status to `pending` if a test result is `positive` but
+    the participant has not yet been contacted (`contacted` != 'yes').
+
+    Returns the modified *scan_data*.
+    """
+    scan_data.loc[scan_data['status_code'].isnull(), 'status_code'] = 'never-tested'
+
+    is_positive = scan_data['status_code'] == 'positive'
+    participant_contacted = scan_data['contacted'] == 'yes'
+
+    scan_data.loc[(is_positive) & ~(participant_contacted), 'status_code'] = 'pending'
+
+    return scan_data
+
+
+if __name__ == '__main__':
+    base = Path(__file__).resolve().parent.parent
+
+    parser = argparse.ArgumentParser(
+        description="Join a REDCAP export csv file with a SCAN return of results ID3C export csv file."
+    )
+    parser.add_argument("redcap_data", help="REDCap records export csv")
+    parser.add_argument("id3c_data", help="SCAN return of results from ID3C export csv", nargs="?", default=sys.stdin)
+    parser.add_argument("output", help="A destination for the output csv", nargs="?",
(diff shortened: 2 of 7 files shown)


bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,69 @@
...
+    Sets the *scan_data* status to `notReceived` in the unlikely event that a
+    REDCap record's barcode is not found in the ID3C samples.
+
+    Sets the *scan_data* status to `pending-other` if a test result is positive
+    (`DET`) but the participant has not yet been contacted (`contacted` == 1.0).
+
+    Returns the modified *scan_data*.
+    """
+    # TODO should we log a warning here?
+    scan_data.loc[scan_data['status_code'].isnull(), 'status_code'] = 'notReceived'
```
kairstenfay: A status of null would mean the REDCap encounter data came in before the sample was created in the database.

bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,69 @@
...
+    base = Path(__file__).resolve().parent.parent
+
+    parser = argparse.ArgumentParser(
+        description="Join a REDCAP export csv file with a SCAN return of results ID3C export csv file.",
+        formatter_class=argparse.RawTextHelpFormatter
+    )
+    parser.add_argument("redcap_data", help="REDCap records export csv")
+    parser.add_argument("id3c_data", help="SCAN return of results from ID3C export csv", nargs="?", default=sys.stdin)
+    parser.add_argument("output", help="A destination for the output csv", nargs="?", default="-")
+    # TODO default is push to S3 once available
```
kairstenfay: I'll replace `default` with the S3 file.

bin/scan-return-of-results/workflow-scan-lab-med
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# usage: workflow-scan-lab-med
+#        workflow-scan-lab-med --help
+#
+# Joins REDCap SCAN data and the SCAN return of results view in ID3C. Parses
+# these data into a format appropriate for UW Lab Med's results portal.
+#
+# Resulting transformed csv is by default stored at :
+#    s3://  TODO
```
kairstenfay: Will update this once we know the URI

kairstenfay:
[Command to return Name + Results if given correct Barcode + DOB](https://trello.com/c/K2uPFEMN/404-command-to-return-name-results-if-given-correct-barcode-dob)

bin/scan-return-of-results/workflow-scan-lab-med
```diff
@@ -0,0 +1,68 @@
...
+
+    export-redcap-data
+    ./export-id3c-scan-return-results > "$datadir/id3c-scan-ror-export.csv"
+
+    ./transform "$datadir/redcap-export.csv" < <(./export-id3c-scan-return-results)
+}
+
+# Exports data using all given REDCap project API tokens, appending to one csv
+export-redcap-data() {
+    local redcap_fields="participant_first_name,participant_last_name,birthday,pre_scan_barcode,contacted"
```
joverlee521: Ditto `return_utm_barcode`

bin/scan-return-of-results/transform
kairstenfay: Ah you're right. 

Jover and I chatted though and agreed that this situation could arise if a bogus barcode is entered in a REDCap project.

kairstenfay: What would be the appropriate code here? `never-tested`, one that [Misja proposed on Slack](https://seattle-flu-study.slack.com/archives/CL1Q38ATG/p1584745444241600?thread_ts=1584740812.239700&cid=CL1Q38ATG)?

joverlee521: Yes? Hopefully this never happens since the barcodes are scanned in?

bin/scan-return-of-results/export-redcap-scan
```diff
@@ -0,0 +1,31 @@
...
+# Does not export field names (csv header).
+#
+set -euo pipefail
+
+: "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
+
+fields="${1:?A comma-separated string of requested fields from REDCap is required as the first argument.}"
+api_token="${2:?A REDCap API token is required as the second argument.}"
+
+filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != '' and [pre_scan_barcode] != ''"
```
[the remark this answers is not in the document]
kairstenfay: Oh that's the barcode field I should be using? Now `pre_scan_barcode`'s name make a lot more sense..

bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,69 @@
...
+    Sets the *scan_data* status to `pending-other` if a test result is positive
+    (`DET`) but the participant has not yet been contacted (`contacted` == 1.0).
+
+    Returns the modified *scan_data*.
+    """
+    # TODO should we log a warning here?
+    scan_data.loc[scan_data['status_code'].isnull(), 'status_code'] = 'notReceived'
+
+    is_positive = scan_data['status_code'] == 'DET'
+    participant_contacted = scan_data['contacted'] == 1.0
```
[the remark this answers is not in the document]
kairstenfay: thank you

bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,67 @@
...
+if __name__ == '__main__':
+    base = Path(__file__).resolve().parent.parent
+
+    parser = argparse.ArgumentParser(
+        description="Join a REDCAP export csv file with a SCAN return of results ID3C export csv file."
+    )
+    parser.add_argument("redcap_data", help="REDCap records export csv")
+    parser.add_argument("id3c_data", help="SCAN return of results from ID3C export csv", nargs="?", default=sys.stdin)
+    parser.add_argument("output", help="A destination for the output csv", nargs="?",
+        default="s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv")
```
kairstenfay: I don't have credentials for this S3 bucket, and writing to this file with pandas quietly fails.

bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,67 @@
...
+    the participant has not yet been contacted (`contacted` != 'yes').
+
+    Returns the modified *scan_data*.
+    """
+    scan_data.loc[scan_data['status_code'].isnull(), 'status_code'] = 'never-tested'
+
+    is_positive = scan_data['status_code'] == 'positive'
+    participant_contacted = scan_data['contacted'] == 'yes'
+
+    scan_data.loc[(is_positive) & ~(participant_contacted), 'status_code'] = 'pending'
```
kairstenfay: When run inside the id3c pipenv environment, I get this warning:
```
/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/ops/array_ops.py:253: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  res_values = method(rvalues)
```

It looks like it [could be a numpy bug](https://stackoverflow.com/a/46721064/6287798)? I do not see the same FutureWarning when using my base python install.

bin/scan-return-of-results/transform
tsibley: quietly fails? ugggh.

bin/scan-return-of-results/workflow-scan-lab-med
```diff
@@ -26,25 +26,18 @@ main() {
...
-    local datadir="$(mktemp -d -t scan-return-of-results-XXXXXX)"
     local redcap_project_tokens=("$REDCAP_API_TOKEN_EN" "$REDCAP_API_TOKEN_ES")
-
-    # Always leave and delete this temporary directory regardless of exit code
-    trap "rm -rf '$datadir'" EXIT
-
-    export-redcap-data
-
-    ./transform "$datadir/redcap-export.csv" < <(./export-id3c-scan-return-results)
+    ./transform <(export-redcap-data) <(./export-id3c-scan-return-results)
```
kairstenfay: Oh all I had to do was put the function in `<()` like the other script?? I swear I tried this earlier and couldn't figure out how to make it work. Awesome.

bin/scan-return-of-results/workflow-scan-lab-med
```diff
@@ -26,19 +26,14 @@ main() {
...
 export-redcap-data() {
-    local redcap_fields="participant_first_name,participant_last_name,birthday,return_utm_barcode,contacted"
-    # Create csv file with headers only
-    echo $redcap_fields
-
-    for token in "${redcap_project_tokens[@]}"; do
-        REDCAP_API_TOKEN="$token" ./export-redcap-scan "$redcap_fields"
-    done;
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_EN" ./export-redcap-scan
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_ES" ./export-redcap-scan | tail -n +2
```
kairstenfay: The trade-off here is easier local development/testing at the cost of repeating all the calls to `./export-redcap-scan`? Seems fine to me.

crontabs/uw-lab-med-return-of-results
```diff
@@ -9,4 +9,4 @@ ENVD=/opt/backoffice/id3c-production/env.d
 
 
 # Export the latest, parsed SCAN return of results data to S3 for UW Lab Med's results portal
-35 * * * * ubuntu /opt/backoffice/bin/scan-return-of-results/workflow-scan-lab-med
+35 * * * * ubuntu /opt/backoffice/bin/scan-return-of-results/workflow-scan-lab-med | envdir $ENVD/securelink/ aws s3 cp - s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv
```
kairstenfay: ah, the first part of this pipe will need additional environment variables

kairstenfay:
> Thanks for gluing all these things together! I ended up doing my review in the form of making actual changes, which seemed easier. Please take a look at the commits I pushed up and make sure it all looks ok you?

Thanks for making these improvements. It looks good to me. I pushed one more commit to add an envdir for REDCap SCAN projects.

If you approve I'll merge and deploy.

kairstenfay:
> If you approve I'll merge and deploy.

Sorry, spoke too soon.

When I run `transform` locally, I get this error:
```
Traceback (most recent call last):
  File "./bin/scan-return-of-results/transform", line 69, in <module>
    joined_data = edit_status_code(joined_data)
  File "./bin/scan-return-of-results/transform", line 44, in edit_status_code
    scan_data.loc[(is_positive) & ~(participant_contacted), 'status_code'] = 'pending'
  File "/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/indexing.py", line 669, in __setitem__
    indexer = self._get_setitem_indexer(key)
  File "/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/indexing.py", line 647, in _get_setitem_indexer
    return self._convert_tuple(key)
  File "/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/indexing.py", line 732, in _convert_tuple
    idx = self._convert_to_indexer(k, axis=i)
  File "/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/indexing.py", line 1725, in _convert_to_indexer
    if com.is_bool_indexer(obj):
  File "/home/kairsten/.local/share/virtualenvs/id3c-customizations-OmmbKeV1/lib/python3.6/site-packages/pandas/core/common.py", line 144, in is_bool_indexer
    raise ValueError(na_msg)
ValueError: cannot mask with array containing NA / NaN values
```

kairstenfay:
> > If you approve I'll merge and deploy.
> 
> Sorry, spoke too soon.
> 
> When I run `transform` locally, I get this error:

This is now fixed with my latest commit.

bin/scan-return-of-results/workflow-scan-lab-med
tsibley: Yep! It's possible you tried `< <(…) < <(…)` which is subtly different from `<(…) <(…)`?

bin/scan-return-of-results/workflow-scan-lab-med
tsibley: and still just 2 lines to add when adding a new project: an existence assertion at the top and a new line in this function (instead of an element in the token array). It would have been possible to keep the loop, but unrolling it seemed to me to clarify the logic to keep only the first output's header.

crontabs/uw-lab-med-return-of-results
tsibley: Ah, yes!

tsibley:
Thanks for the updates!

I didn't get that error from `transform` when running locally. I wonder if this is a different in test data that we're feeding to it?

For the crontab, just realized the command will need to be run inside a Python environment with pandas. id3c-production is fine, I suppose?

kairstenfay:
> I didn't get that error from `transform` when running locally. I wonder if this is a different in test data that we're feeding to it?

Yeah, I had some test data I created with null values for `contacted`. 
 
> For the crontab, just realized the command will need to be run inside a Python environment with pandas. id3c-production is fine, I suppose?

Oh, yup. I just pushed a new commit that uses the id3c-production pipenv environment.

crontabs/uw-lab-med-return-of-results
```diff
@@ -9,4 +12,4 @@ ENVD=/opt/backoffice/id3c-production/env.d
 
 
 # Export the latest, parsed SCAN return of results data to S3 for UW Lab Med's results portal
-35 * * * * ubuntu envdir $ENVD/redcap-scan/ /opt/backoffice/bin/scan-return-of-results/workflow-scan-lab-med | envdir $ENVD/securelink/ aws s3 cp - s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv
+35 * * * * pipenv run ubuntu envdir $ENVD/redcap-scan/ /opt/backoffice/bin/scan-return-of-results/workflow-scan-lab-med | envdir $ENVD/securelink/ aws s3 cp - s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv
```
tsibley: `ubuntu` isn't part of the command; it's the user to run as and is a separate field in the crontab.

kairstenfay: Oh yeah, I got it turned around. It should be `ubuntu pipenv run ...`?

tsibley: Yep.

kairstenfay: Fixed up in the latest commit.

bin/scan-return-of-results/workflow-scan-lab-med
kairstenfay: Yes I think that was it.

bin/scan-return-of-results/workflow-scan-lab-med
kairstenfay: Yeah, I value clarity above cleverness anyway, so this is a sensible two extra lines per project.

bin/scan-return-of-results/transform
```diff
@@ -0,0 +1,70 @@
...
+    args = parser.parse_args()
+
+    redcap_data = parse_redcap(args.redcap_data)
+
+    id3c_data = pd.read_csv(args.id3c_data, dtype = 'string')
+    id3c_data['qrcode'] = id3c_data['qrcode'].str.lower().str.strip()
+
+    # We can't return results without participant info so keep all from redcap_data
+    joined_data = redcap_data.merge(id3c_data, how='left', on='qrcode')
+    joined_data = edit_status_code(joined_data)
```
tsibley: When I was quickly testing this again, I saw some example data from Misja in the output with bogus barcodes. It seems to me that we wouldn't want to include bogus barcodes? That is, that we'd want to make this an inner join, or maybe a right join (given the current LHS and RHS)?

kairstenfay: Then we would generate no `never-tested` codes. Do we want to account for the (hopefully rare) occasion an incorrect barcode is scanned into REDCap?

tsibley: My thinking: We pull the barcode from REDCap that was scanned by the receiving team, right? If that barcode is invalid, then how would a participant even find their result given the correct barcode they presumably have + their DOB? And if they also somehow used the same incorrect barcode, it seems more accurate and more useful to show the participant a "we don't recognize that barcode" error instead of a "won't/can't be tested" result. This would presumably lead to a call/email to the participant support contacts and resolution of the issue one way or another.

I agree that changing the join means not generating `never-tested` for now. We will generate those statues in the future though, given additional data to decide when it is appropriate.

kairstenfay: That sound fine to me. I wasn't sure if we wanted to account for invalid barcodes, so I by default included them. I'm OK with excluding them using an inner join here.

kairstenfay: @tsibley pushed in a new commit

kairstenfay:
> Removed the Spanish project for now as it looks like it isn't in production yet.

Good call.

> Since there are not any records in REDCap yet and this pipeline fails when that's the case, maybe let's hold off on deploying this until there are records? We could catch [pandas.errors.EmptyDataError](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.errors.EmptyDataError.html) for now and let it slide, but down the road that'd suppress actual data generation errors we'd want to see.

Ah, fair point. I'll wait until ... tomorrow? :crossed_fingers:

tsibley:
@seattleflu/id3c-devs I just pushed two additional commits, please take a look. If they're good, then I think this is ok to merge.

kairstenfay:
@tsibley Looks good to me. I pushed one more commit to trim white space from participants' names.

joverlee521:
Looks good to me!

PR #28, opened 2020-03-23
Tear down the now-unused id3c-restricted deployment

tsibley:
A short-lived deployment which enabled us to receive sensitive hCoV-19
results from the lab during an interim period.  We now receive that data
into our standard production deployment with more granular access
controls.


tsibley:
[Tear down id3c-restricted](https://trello.com/c/7yldwrCG/426-tear-down-id3c-restricted)

PR #29, opened 2020-03-24
crontabs: reportable conditions

joverlee521:
Run 5 minutes after presence/absence results are uploaded to check
if any contain reportable conditions.

### DIFF
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -30,6 +30,10 @@ ENVD=/opt/backoffice/id3c-production/env.d
 # Test results are pushed at arbitrary times now, so check for new ones every 15m.
 */15 * * * * ubuntu pipenv run id3c etl presence-absence --commit
 
+# 5 minutes after presence/absence results are uploaded,
+# check if any contain reportable conditions.
+5-50/15 * * * * ubuntu envdir $ENVD/slack pipenv run id3c reportable-conditions notify --commit
+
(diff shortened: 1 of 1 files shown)


joverlee521:
[Turn on reportable conditions cronjob](https://trello.com/c/kxzaNovX/433-turn-on-reportable-conditions-cronjob)

crontabs/id3c-production
```diff
@@ -30,6 +30,10 @@ ENVD=/opt/backoffice/id3c-production/env.d
 # Test results are pushed at arbitrary times now, so check for new ones every 15m.
 */15 * * * * ubuntu pipenv run id3c etl presence-absence --commit
 
+# 5 minutes after presence/absence results are uploaded,
+# check if any contain reportable conditions.
+5-50/15 * * * * ubuntu envdir $ENVD/slack pipenv run id3c reportable-conditions notify --commit
```
kairstenfay: I just realized that in my recent backoffice PR for SCAN RoR, I invoked my script with `ubuntu pipenv run envdir $ENVD/...` instead of `ubuntu envidr $ENVD/... pipenv run`.

Are either acceptable? Do we prefer one over the other?

tsibley: Either is acceptable. :-) As long as they're not modifying the same environment variables (which they aren't), they're equivalent.

PR #31, opened 2020-03-25
Generate lab result report PDFs for SCAN 

tsibley:
This is the backoffice cronjob to generate PDFs. Most of the actual PDF generation code lives in https://github.com/seattleflu/lab-result-reports.

### DIFF
--- a/bin/scan-return-of-results/generate-pdfs
+++ b/bin/scan-return-of-results/generate-pdfs
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${S3_PREFIX:=s3://dokku-stack-phi/covid19/scan-study-inbound}"
+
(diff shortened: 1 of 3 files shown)


tsibley:
[Generate PDFs from Lab Med S3 file for SCAN RoR](https://trello.com/c/Hg63Je5c/410-generate-pdfs-from-lab-med-s3-file-for-scan-ror)

tsibley:
> Will this generate a new PDF for every participant every time?

Yeah. My thinking was that this is much simpler than doing a diff and only regenerating those. It should be tenable for the near future, and maybe quite a while longer. I planned to keep an eye on it and see at what point we'd want to switch to incremental generation. Worth noting too that data transfer into S3 is free, so it won't incur those extra costs for Lab Med.

tsibley:
(Updated the commit message to include that explanation.)

PR #34, opened 2020-03-31
scan-return-of-results: Don't make PDFs on "never-tested"

kairstenfay:
Add "never-tested" to the list of excluded barcodes in PDF generation.


kairstenfay:
[Use `able_to_test` as `never-tested` in SCAN RoR view](https://trello.com/c/Qf6GyDm3/465-use-abletotest-as-never-tested-in-scan-ror-view)

kairstenfay:
I incorrectly assumed we weren't returning results to people "never-tested".

PR #35, opened 2020-03-31
scan-return-of-results: Mask 'inconclusive' when not contacted

kairstenfay:
Mask 'inconclusive' results as 'pending' (similar to what we do for
positives) when a participant has not yet been contacted by study
member.


kairstenfay:
[mask inconclusive SCAN RoR results for participants not yet contacted](https://trello.com/c/8fZSd0wZ/488-mask-inconclusive-scan-ror-results-for-participants-not-yet-contacted)

PR #36, opened 2020-04-03
Add import-data-dictionary-to-redcap script.

joverlee521:
1. Exports data dictionary from REDCap project.
2. Injects a span tag with provided HTML attributes on `section_header`,
`field_label`, `select_choices_or_calculations` and `field_note`
of listed field names.
3. If `--import-to-redcap` flag is provided, import the new data
dictionary to REDCap project.


joverlee521:
[Investigate how to bulk add language tags to REDCap projects](https://trello.com/c/kGNLvRed/479-investigate-how-to-bulk-add-language-tags-to-redcap-projects)

kairstenfay:
Thanks for tackling this, Jover. A few thoughts as I continue to look over this script:

- I feel like we may want to split this python script into scripts: one that does the data dictionary export from REDCap, and one that does the transform/import to REDCap. One thing that made this idea stand out to me here was the dual output, i.e. double printing to stdout by default. I can not imagine a scenario where we would want both the original REDCap dictionary and the new one printed to stdout serially. 

- I like that you used balanced words here to describe what this script does (export/import). However, using the phrase "import" when we mean uploading data to REDCap could produce some confusion. I'm wondering if the words "download/upload" could be more appropriate here. 

Curious to know your thoughts.

joverlee521:
> Thanks for tackling this, Jover. A few thoughts as I continue to look over this script:
> 
> * I feel like we may want to split this python script into scripts: one that does the data dictionary export from REDCap, and one that does the transform/import to REDCap. One thing that made this idea stand out to me here was the dual output, i.e. double printing to stdout by default. I can not imagine a scenario where we would want both the original REDCap dictionary and the new one printed to stdout serially.
> * I like that you used balanced words here to describe what this script does (export/import). However, using the phrase "import" when we mean uploading data to REDCap could produce some confusion. I'm wondering if the words "download/upload" could be more appropriate here.
> 
> Curious to know your thoughts.

1. I'm fine with splitting the scripts. We can have a redcap directory like `scan-return-of-results`

2. I used export/import to mimic the REDCap API [docs](https://redcap.iths.org/redcap_v9.7.2/API/help.php?content=exp_metadata), I have no particular attachment to the naming there.

kairstenfay:
Rebased onto master plus added one bug fix commit

kairstenfay:
Thanks so much @joverlee521 for doing the bulk of this work especially around formatting REDCap API requests and doing the logic for injecting HTML tags into data dictionaries.

I made a few changes, which are now ready for review:
1. split the main "import" script into two scripts: download and upload
2. added some logging when POSTing to the REDCap API. It's a little wonky because it only accepts "debug" or a default of "info" for log level, but perhaps it's good enough? 
3. Added a "batch" script that can be extended to submitting multiple data dictionaries for multiple (language) REDCap projects. I expect the structure of this script will change as we add more projects. For example, we will probably want to export the SCAN-EN data dictionary only once then have other "upload" processes read from that output (rather than injecting the "download" script and its output each time). 

I welcome feedback anywhere, especially on 2 and 3.

bin/redcap-data-dictionary/post_to_redcap.py
```diff
@@ -5,7 +5,8 @@
 import logging
 from typing import Any, Dict
 
-logging.basicConfig(level=logging.INFO)
+LOG_LEVEL = os.environ.get('LOG_LEVEL')
+logging.basicConfig(level=logging.DEBUG if LOG_LEVEL == 'debug' else logging.INFO)
```
kairstenfay, line 9: I see now in the `lab-result-reports` repo a better example of how to do this:
```
LOG_LEVEL = os.environ.get("LOG_LEVEL", "debug").upper()


logging.basicConfig(
    level = logging.ERROR,
    format = "[%(asctime)s] %(levelname)-8s %(message)s",
    datefmt = "%Y-%m-%d %H:%M:%S%z",
    stream = stderr)

logging.captureWarnings(True)

log = logging.getLogger(__name__)
log.setLevel(LOG_LEVEL)
```

joverlee521:
> Thanks so much @joverlee521 for doing the bulk of this work especially around formatting REDCap API requests and doing the logic for injecting HTML tags into data dictionaries.
> 
> I made a few changes, which are now ready for review:
> 
> 1. split the main "import" script into two scripts: download and upload
> 2. added some logging when POSTing to the REDCap API. It's a little wonky because it only accepts "debug" or a default of "info" for log level, but perhaps it's good enough?
> 3. Added a "batch" script that can be extended to submitting multiple data dictionaries for multiple (language) REDCap projects. I expect the structure of this script will change as we add more projects. For example, we will probably want to export the SCAN-EN data dictionary only once then have other "upload" processes read from that output (rather than injecting the "download" script and its output each time).
> 
> I welcome feedback anywhere, especially on 2 and 3.

For 3, I don't think we want to export the SCAN-EN data dictionary. These projects will be already translated, so we need to keep the translations from each project. So the flow would be to download the translated project data dictionary, inject the tags and then upload the new data dictionary back to the same project.

kairstenfay:
> > Thanks so much @joverlee521 for doing the bulk of this work especially around formatting REDCap API requests and doing the logic for injecting HTML tags into data dictionaries.
> > I made a few changes, which are now ready for review:
> > 
> > 1. split the main "import" script into two scripts: download and upload
> > 2. added some logging when POSTing to the REDCap API. It's a little wonky because it only accepts "debug" or a default of "info" for log level, but perhaps it's good enough?
> > 3. Added a "batch" script that can be extended to submitting multiple data dictionaries for multiple (language) REDCap projects. I expect the structure of this script will change as we add more projects. For example, we will probably want to export the SCAN-EN data dictionary only once then have other "upload" processes read from that output (rather than injecting the "download" script and its output each time).
> > 
> > I welcome feedback anywhere, especially on 2 and 3.
> 
> For 3, I don't think we want to export the SCAN-EN data dictionary. These projects will be already translated, so we need to keep the translations from each project. So the flow would be to download the translated project data dictionary, inject the tags and then upload the new data dictionary back to the same project.
> 
> Edit: This got me thinking we probably don't need the batch script. My thought is we would do the html injection as a last thing before the project goes into production so we include all the correct translations. This means the timing for using this script on each project will probably be different.

Good point... for some reason I thought translations would be done _after_ we injected the html attributes. But you're right. Each project would only want this in-place modification to be run once, and the timing will differ based on the project. I should just remove the batch script.

kairstenfay:
@joverlee521 I don't seem to be able to request your review on this (since you created the PR). Would you mind taking a look at everything and letting me know if anything looks out of place?

joverlee521:
@kairstenfay Looks good to me! (And it clearly worked very well on the TEST REDcap project)

PR #37, opened 2020-04-06
SCAN: Rely only on the barcode scanned during returned kit processing

tsibley:
"return_utm_barcode" is the barcode which is most reliable and has the
final say in the case of kit mixups.  Switching to this avoids having to
manually mask data entry or pre-shipment scanning errors.

Duplicates in this field are a mistake and should be very rare, but drop
any that are seen and issue a warning.  Manual followup with the lab
team is required in such cases.

As a result of not using the pre-shipping or participant-entered
barcodes, rows where status_code == not-received will drop out from the
SecureLink data file, which will land more people on the SecureLink SCAN
error page.  The error page already says it may not be received yet, and
I've added another PR to make it a bit better still.


tsibley:
[Barcode handling for SCAN return of results](https://trello.com/c/2879l8mv/511-barcode-handling-for-scan-return-of-results)

tsibley:
Related SecureLink PR https://github.com/nkrumm/securelink/pull/9

PR #38, opened 2020-04-17
Generate PDFs for es, vi, zh-Hans

kairstenfay:
Generate PDFs for three additional finalized languages: Spanish,
Vietnamese, and Chinese (simplified).

Use the latest Docker image for PDF generation.

### DIFF
--- a/bin/scan-return-of-results/generate-pdfs
+++ b/bin/scan-return-of-results/generate-pdfs
@@ -5,7 +5,7 @@ params="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results
 output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}}"
 
 for lang in "en" "es" "vi" "zh-Hans"; do
-    echo "Generating PDFs for language code «$lang»"
+    echo "Generating PDFs for language code «${lang}»"
 
     docker run \
         --rm \
--- a/bin/scan-return-of-results/generate-pdfs
+++ b/bin/scan-return-of-results/generate-pdfs
@@ -2,10 +2,10 @@
 set -euo pipefail
 
 params="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
+output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}}"
 
 for lang in "en" "es" "vi" "zh-Hans"; do
     echo "Generating PDFs for language code «$lang»"
-    output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}-$lang.pdf}"
 
     docker run \
         --rm \
@@ -17,6 +17,6 @@ for lang in "en" "es" "vi" "zh-Hans"; do
             fill-template \
                 --template "scan/report-$lang.tex" \
                 --params "$params" \
-                --output "$output" \
+                --output "$output-$lang.pdf" \
                 --filter 'status_code not in ["not-received", "pending"]'
 done;
--- a/bin/scan-return-of-results/generate-pdfs
+++ b/bin/scan-return-of-results/generate-pdfs
@@ -2,17 +2,21 @@
 set -euo pipefail
 
 params="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
-output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}-en.pdf}"
 
-docker run \
-    --rm \
-    --interactive \
-    --env=AWS_ACCESS_KEY_ID \
-    --env=AWS_SECRET_ACCESS_KEY \
-    --env=LOG_LEVEL \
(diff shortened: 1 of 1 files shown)


bin/scan-return-of-results/generate-pdfs
```diff
@@ -2,17 +2,21 @@
...
-    --env=LOG_LEVEL \
-    seattleflu/lab-result-reports:build-8 \
-        fill-template \
-            --template scan/report-en.tex \
-            --params "$params" \
-            --output "$output" \
-            --filter 'status_code not in ["not-received", "pending"]'
+for lang in "en" "es" "vi" "zh-Hans"; do
+    echo "Generating PDFs for language code «$lang»"
+    output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}-$lang.pdf}"
```
kairstenfay, line 8: Should the declaration of the `output` variable here use `local` as a prefix?

kairstenfay:
[Make backoffice job generate PDFs for all languages](https://trello.com/c/NC9m4sgx/532-make-backoffice-job-generate-pdfs-for-all-languages)

bin/scan-return-of-results/generate-pdfs
```diff
@@ -2,17 +2,21 @@
...
-    --env=AWS_SECRET_ACCESS_KEY \
-    --env=LOG_LEVEL \
-    seattleflu/lab-result-reports:build-8 \
-        fill-template \
-            --template scan/report-en.tex \
-            --params "$params" \
-            --output "$output" \
-            --filter 'status_code not in ["not-received", "pending"]'
+for lang in "en" "es" "vi" "zh-Hans"; do
+    echo "Generating PDFs for language code «$lang»"
```
joverlee521, line 7: I get the following error when I try to run the script locally: 
```
./bin/scan-return-of-results/generate-pdfs: line 7: lang�: unbound variable
```
I think `$lang` needs to be out of the double quotes? Something like:
```
echo "Generating PDFs for language code «" $lang "»"
```

kairstenfay: In bash, we typically want to enclose string variables with double quotes, so I don't think the answer is to remove `$lang` from the double quotes. I thought I had done the string interpolation correctly, but now I'm confused why it wouldn't work on your machine.

joverlee521: I'll attribute this to something weird with MACs 😆

bin/scan-return-of-results/generate-pdfs
tsibley: Nope, because `local` is function-scoped:

> Local variables can only be used within a function; they are visible
> only to the function where they are defined and its children.

(from `help local`)

bin/scan-return-of-results/generate-pdfs
tsibley: This looks to me like an old version of Bash (like the ancient version shipped on macOS) includes the `»` in the variable name when tokenizing. Since `lang»` isn't a variable name that exists, it errors (because of `set -u`). I expect that using the unambiguous variable interpolation form would help:

```
echo "Generating PDFs for language code «${lang}»"
```

@joverlee521 To confirm, you could quickly try:

```
foo=bar
echo "Hello, «$foo»"
echo "Hello, «${foo}»"
```

bin/scan-return-of-results/generate-pdfs
tsibley, line 8: This makes the program ~impossible to manually test in dev because there's no way to include `$lang` when you pass in a second parameter (`$2`). That is, I used to test locally like this:

```
./generate-pdfs results.csv "{qrcode}.pdf"
```

but now each language would overwrite the last, leaving me with just Chinese PDFs.

Here's one idea:

```
output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}}"
…
        --output "$output-$lang.pdf" \
```

bin/scan-return-of-results/generate-pdfs
joverlee521: Yup, `echo "Hello, «${foo}»"` worked for me.

bin/scan-return-of-results/generate-pdfs
kairstenfay: Ah yes I see that now. I did not recognize the `${2:-filepath}` syntax as accepting a parameter while specifying a default even though I've used that before myself. 
There's still a ways to go in familiarizing myself with Bash.

Fixed in the latest commit.

bin/scan-return-of-results/generate-pdfs
kairstenfay: I'm not sure if the unambiguous variable interpolation needs to be done for every instance of `$lang` in the script... @joverlee521 would you mind trying to run the script now after my latest commit?

joverlee521: @kairstenfay Everything ran as expected after your latest commit 🎉

tsibley: > I'm not sure if the unambiguous variable interpolation needs to be done for every instance of `$lang` in the script...

Not necessary. It's not tied to `$lang` itself, but on the place the interpolation occurs and the immediately following character.

PR #39, opened 2020-04-23
Optimize pdf gen

kairstenfay:
Generate SCAN results PDFs only for new or changed data.

### DIFF
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -1,8 +1,58 @@
 #!/bin/bash
+# usage: ./bin/generate [results] [latest_results]
+#
+# Updates the Securelink S3 results CSV and generates PDFs for new or changed
+# results.
+#
+# *results* and *latest_results* are local or S3 filenames of CSVs exported by
+# the `./generate-results-csv` script.
+#
+# If *results* is not provided, defaults to reading from the remote results CSV
+# at the Securelink S3 bucket.
+#
+# If *latest_results* is not provided, generates these data on the fly using
+# live data from ID3C and REDCap.
+#
 set -euo pipefail
 
 bin="$(dirname "$0")"
-results="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
 
-"$bin"/generate-results-csv | aws s3 cp - "$results"
-"$bin"/generate-pdfs "$results"
+# Relative path to our containing repo, accessible by our Docker container
+base="$(dirname "$0")/../.."
+
+results="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
+latest_results="${2:-}"
+datadir="$(mktemp -d -t -p "$base" scan-return-results-XXXXXX)"
+
+# Always leave and delete this temporary directory regardless of exit code
+trap "rm -rf '$datadir'" EXIT
+
+# Copy the existing results file into the temp directory accessible by Docker
+if [[ "$results" == "s3://"* ]]; then
+    echo "Copying S3 results for comparison"
+    aws s3 cp "$results" "$datadir/previous_results.csv"
+else
+    echo "Copying local results for comparison"
+    cp "$results" "$datadir/previous_results.csv"
+fi
+
+# If latest results not provided, generate them now
+if [[ -z "$latest_results" ]]; then
+    "$bin"/generate-results-csv > "$datadir/latest_results.csv"
+else
+    cp "$latest_results" "$datadir/latest_results.csv"
+fi
+
+# Construct a CSV containing the first row (column names) of the results file
+# and new or changed rows from the previous version
+head -n 1 "$datadir/previous_results.csv" > "$datadir/diff.csv"
+grep -vf "$datadir/previous_results.csv" "$datadir/latest_results.csv" >> "$datadir/diff.csv"
+
+
+"$bin"/generate-pdfs "$datadir/diff.csv"
+
+
+# Upload latest results to S3
+if [[ "$results" == "s3://"* ]]; then
+    aws s3 cp "$datadir/latest_results.csv" "$results"
+fi
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -1,16 +1,16 @@
 #!/bin/bash
-# usage: ./bin/generate [results] [latest_results]
+# usage: ./bin/generate [<results>] [<latest_results>]
 #
 # Updates the Securelink S3 results CSV and generates PDFs for new or changed
 # results.
 #
-# *results* and *latest_results* are local or S3 filenames of CSVs exported by
+# <results> and <latest_results> are local or S3 filenames of CSVs exported by
 # the `./generate-results-csv` script.
 #
-# If *results* is not provided, defaults to reading from the remote results CSV
+# If <results> is not provided, defaults to reading from the remote results CSV
 # at the Securelink S3 bucket.
 #
-# If *latest_results* is not provided, generates these data on the fly using
+# If <latest_results> is not provided, generates these data on the fly using
 # live data from ID3C and REDCap.
 #
 set -euo pipefail
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -58,7 +58,10 @@ main() {
     # Construct a CSV containing the first row (column names) of the results file
     # and new or changed rows from the previous version
     head -n 1 "$datadir/latest_results.csv" > "$datadir/diff.csv"
-    grep -vf "$datadir/previous_results.csv" "$datadir/latest_results.csv" >> "$datadir/diff.csv"
+    comm -13 \
+        <(tail -n +2 "$datadir/previous_results.csv" | sort) \
+        <(tail -n +2 "$datadir/latest_results.csv" | sort) \
+        >> "$datadir/diff.csv"
 
 
     "$bin"/generate-pdfs "$datadir/diff.csv"
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -57,7 +57,7 @@ main() {
 
     # Construct a CSV containing the first row (column names) of the results file
     # and new or changed rows from the previous version
-    head -n 1 "$datadir/previous_results.csv" > "$datadir/diff.csv"
+    head -n 1 "$datadir/latest_results.csv" > "$datadir/diff.csv"
     grep -vf "$datadir/previous_results.csv" "$datadir/latest_results.csv" >> "$datadir/diff.csv"
 
 
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -24,6 +24,16 @@ results="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_result
 latest_results="${2:-}"
 
 main() {
+    for arg; do
+        case "$arg" in
+            -h|--help)
+                print-help
+                exit
+                ;;
+            *)
+        esac
+    done
+
     datadir="$(mktemp -d -t -p "$base" scan-return-results-XXXXXX)"
 
     # Always leave and delete this temporary directory regardless of exit code
@@ -60,5 +70,12 @@ main() {
     fi
 }
 
+print-help() {
+    # Print the help comments at the top of this file ($0)
+    perl -MEnglish -ne '
+        s/^# ?// or exit;
+        print if $INPUT_LINE_NUMBER >= 2;
+    ' "$0"
+}
 
 main "$@"
--- a/bin/scan-return-of-results/generate
+++ b/bin/scan-return-of-results/generate
@@ -34,7 +34,7 @@ main() {
         esac
     done
 
-    datadir="$(mktemp -d -t -p "$base" scan-return-results-XXXXXX)"
+    datadir="$(TMPDIR="$base" mktemp -d -t scan-return-results-XXXXXX)"
 
     # Always leave and delete this temporary directory regardless of exit code
     trap "rm -rf '$datadir'" EXIT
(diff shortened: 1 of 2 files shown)


bin/scan-return-of-results/generate-pdfs
```diff
@@ -8,15 +8,15 @@ for lang in "en" "es" "vi" "zh-Hans"; do
...
+        --rm --volume $PWD:/cwd:ro \
         --interactive \
         --env=AWS_ACCESS_KEY_ID \
         --env=AWS_SECRET_ACCESS_KEY \
         --env=LOG_LEVEL \
         seattleflu/lab-result-reports:build-10 \
             fill-template \
                 --template "scan/report-$lang.tex" \
-                --params "$params" \
+                --params "/cwd/$params" \
```
kairstenfay, line 19: Not sure if we want to hard code this prefix to `$params` here or make the user provide it.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,58 @@
...
+#
+# *results* and *latest_results* are local or S3 filenames of CSVs exported by
+# the `./generate-results-csv` script.
+#
+# If *results* is not provided, defaults to reading from the remote results CSV
+# at the Securelink S3 bucket.
+#
+# If *latest_results* is not provided, generates these data on the fly using
+# live data from ID3C and REDCap.
+#
```
kairstenfay, line 15: These lines are long, making the output of `--help` unwieldy. Is there a better way to format the help message other than to reduce the line lengths?

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,58 @@
...
 set -euo pipefail
 
 bin="$(dirname "$0")"
+
+# Relative path to our containing repo, accessible by our Docker container
+base="$(dirname "$0")/../.."
```
kairstenfay, line 21: I am not convinced this is the best way to do this.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,58 @@
...
+
 results="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
+latest_results="${2:-}"
+datadir="$(mktemp -d -t -p "$base" scan-return-results-XXXXXX)"
+
+# Always leave and delete this temporary directory regardless of exit code
+trap "rm -rf '$datadir'" EXIT
+
+# Copy the existing results file into the temp directory accessible by Docker
+if [[ "$results" == "s3://"* ]]; then
```
kairstenfay: Is matching the start of the string an advisable way to determine if a file is an S3 file or not?

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,58 @@
...
+#
 set -euo pipefail
 
 bin="$(dirname "$0")"
+
+# Relative path to our containing repo, accessible by our Docker container
+base="$(dirname "$0")/../.."
+
 results="${1:-s3://dokku-stack-phi/covid19/scan-study-inbound/scan_return_results.csv}"
+latest_results="${2:-}"
```
kairstenfay, line 24: In local development, if you just want to specify `$latest_results` but let `$results` remain as the default, it leads to a poor user experience. I'm wondering if these parameters would be better of as options instead of positional arguments.

kairstenfay:
[Generate PDFs only for changed/new results data](https://trello.com/c/Fl88AxCN/539-generate-pdfs-only-for-changed-new-results-data)

joverlee521:
I haven't gone through the whole PR yet, but just had a thought: 
Do we need to create new PDFs for old results in new languages as they become available? A diff of the `scan-return-results.csv` will not generate PDFs for new languages of old results.

kairstenfay:
> I haven't gone through the whole PR yet, but just had a thought:
> Do we need to create new PDFs for old results in new languages as they become available? A diff of the `scan-return-results.csv` will not generate PDFs for new languages of old results.

That's an excellent point I had not considered. I will have to think about how to check for new languages.

joverlee521:
> > I haven't gone through the whole PR yet, but just had a thought:
> > Do we need to create new PDFs for old results in new languages as they become available? A diff of the `scan-return-results.csv` will not generate PDFs for new languages of old results.
> 
> That's an excellent point I had not considered. I will have to think about how to check for new languages.

One easy (brute-force) way around this is to do a one-time manual PDF generation of all results for each new language we add.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,81 @@
...
+
+
+    "$bin"/generate-pdfs "$datadir/diff.csv"
+
+
+    # Upload latest results to S3
+    if [[ "$results" == "s3://"* ]]; then
+        aws s3 cp "$datadir/latest_results.csv" "$results"
+    fi
+}
```
kairstenfay, line 71: Big oof here:
Testing this locally, I accidentally overwrote the results on S3 with the `$latest_results` argument I provided manually. 

There should be better guardrails around this, or we should remove the optional `$latest_results` argument altogether (making local development more annoying but production safer).

kairstenfay: Or we can just be sure to re-run the script without passing `$latest_results`, restoring the file on S3 to its previous condition.

bin/scan-return-of-results/generate-pdfs
tsibley: If we hardcode that prefix, then the default s3 url as used in production will no longer work. So the user must provide `/cwd`, or we could detect a relative file path and auto-prefix only in that case, but probably not worth the extra complexity. Dev/testing will be the only time we'll pass in a non-default value, and it's reasonable to expect us to know details of how the program works.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,58 @@
...
+    cp "$latest_results" "$datadir/latest_results.csv"
+fi
+
+# Construct a CSV containing the first row (column names) of the results file
+# and new or changed rows from the previous version
+head -n 1 "$datadir/previous_results.csv" > "$datadir/diff.csv"
+grep -vf "$datadir/previous_results.csv" "$datadir/latest_results.csv" >> "$datadir/diff.csv"
+
+
+"$bin"/generate-pdfs "$datadir/diff.csv"
```
joverlee521: Currently we upload the results to AWS before starting the PDF generation, do we still want to keep that order here or did you switch the order on purpose?

kairstenfay: Ah, I switched it to the end before I started creating a separate `diff.csv` file. 

The upload to AWS should happen before PDF generation since PDF generation takes so long.

bin/scan-return-of-results/generate
tsibley: Unwieldy in what way?

72-80 char lines are considered standard by long-standing convention.

As an aside, the `*arg*` convention for marking up arguments comes from Python's docstrings and ReStructuredText. The convention for command line tools is:

1. In the usage line: `<foo>` for a required positional argument and `[<bar>]` for an optional one.
2. In the rest of the help output, arguments (both required and optional) are referred like `<foo>` and `<bar>`.

bin/scan-return-of-results/generate
tsibley: That's fine, or you can fully-resolve the path (i.e. remove the `..` from the value of `base`) by doing something like:

```
base="$(cd "$(dirname "$0")"/../..; pwd)"
```

bin/scan-return-of-results/generate
tsibley: This pattern looks good to me. The right-hand side doesn't need the quotes, FWIW.

bin/scan-return-of-results/generate
tsibley: Ah, I think we did CSV then PDFs just because the PDFs were implemented later.

PDFs, then CSV makes more sense and will guarantee that all the PDFs are available by the time someone is able to lookup their result.

tsibley:
> One easy (brute-force) way around this is to do a one-time manual PDF generation of all results for each new language we add.

This is my recommendation. We can make it a bit less time if we just generate the PDFs for the new languages and not the existing languages.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,81 @@
...
+        case "$arg" in
+            -h|--help)
+                print-help
+                exit
+                ;;
+            *)
+        esac
+    done
+
+    datadir="$(mktemp -d -t -p "$base" scan-return-results-XXXXXX)"
```
tsibley, line 37: Unfortunately, I believe this isn't a portable invocation of mktemp. The [GNU coreutils mktemp](http://man7.org/linux/man-pages/man1/mktemp.1.html) is not entirely compatible with the [BSD (macOS) mktemp](https://www.freebsd.org/cgi/man.cgi?query=mktemp&sektion=1). In particular, the latter doesn't support `-p` and defines `-t` as requiring a prefix.

For a portable replacement, I suggest:

```
TMPDIR="$base" mktemp -d -t scan-return-results-XXXXXX
```

This is interpreted differently, but with equivalent behaviour, by both GNU and BSD platforms.

GNU coreutils mktemp interprets this as the options `-d` and `-t` plus the optional positional `[<template>]` argument `scan-return-results-XXXXXX`.

BSD mktemp interprets this as the options `-d` and `-t <prefix>`, with no positional `<template>` argument.

Both interpret `-t` as meaning that `<template>` or `<prefix>` should be relative to `TMPDIR` (or `/tmp` if `TMPDIR` is unset).

bin/scan-return-of-results/generate
kairstenfay: I considered that, but I need the `base` to be relative to the current working directory, not absolute.

bin/scan-return-of-results/generate
kairstenfay: Good point, we'd rather delay results than end up with 404s.

bin/scan-return-of-results/generate
kairstenfay: > Unwieldy in what way?
> 
> 72-80 char lines are considered standard by long-standing convention.

The line wrapping looks bad when I make my terminal window narrow.
image

> As an aside, the `*arg*` convention for marking up arguments comes from Python's docstrings and ReStructuredText. The convention for command line tools is:
> 
>     1. In the usage line: `<foo>` for a required positional argument and `[<bar>]` for an optional one.
> 
>     2. In the rest of the help output, arguments (both required and optional) are referred like `<foo>` and `<bar>`.

Ah, thank you. I was being lazy and didn't look it up myself :disappointed:

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,81 @@
...
+    if [[ -z "$latest_results" ]]; then
+        "$bin"/generate-results-csv > "$datadir/latest_results.csv"
+    else
+        cp "$latest_results" "$datadir/latest_results.csv"
+    fi
+
+    # Construct a CSV containing the first row (column names) of the results file
+    # and new or changed rows from the previous version
+    head -n 1 "$datadir/previous_results.csv" > "$datadir/diff.csv"
+    grep -vf "$datadir/previous_results.csv" "$datadir/latest_results.csv" >> "$datadir/diff.csv"
```
tsibley, line 61: Since we're taking rows from the latest results, I think safer to take the header from it as well. If the header changes, we don't want to keep the old one. Additionally, we'll want to diff these files without their headers so as not to duplicate the header line if it changed.

Grep will interpret lines in the file as patterns instead of fixed strings. There's a `-F / --fixed-strings` option we could use that works well in GNU grep… but years ago I recall running into a very hard to troubleshoot bug that ultimately stemmed from a reproducible bug in BSD grep when combining the three options `-v`, `-f`, and `-F`. :-/

Instead, you can achieve the same result with [comm](http://man7.org/linux/man-pages/man1/comm.1.html) (for "common"):

```
head -n 1 "$datadir/latest_results.csv" > "$datadir/diff.csv"
comm -13 \
    <(tail -n +2 "$datadir/previous_results.csv" | sort) \
    <(tail -n +2 "$datadir/latest_results.csv" | sort) \
    >> "$datadir/diff.csv"
```

The `sort`s are because `comm` requires the input to be sorted.

bin/scan-return-of-results/generate-pdfs
tsibley: I have an idea here to make it a bit better. Realized we could use volumes more precisely. Will push/paste somewhere soon.

bin/scan-return-of-results/generate
tsibley: > The line wrapping looks bad when I make my terminal window narrow.

Ah, yeah. The only way around this is to dynamically rewrap the text based on `$COLS`, which `argparse` and `click` will do, but is maybe a bit too much hassle for a short shell script like this...

kairstenfay: Great, just wanted to make sure I wasn't missing any easy fix. I'll let it go for now.

bin/scan-return-of-results/generate
```diff
@@ -1,8 +1,81 @@
...
+
+
+    "$bin"/generate-pdfs "$datadir/diff.csv"
```
kairstenfay, line 64: I'm realizing now that we probably want to provide an optional `output` argument that can be passed to the `generate-pdfs` script so the PDFs can be written locally.

bin/scan-return-of-results/generate
kairstenfay: Addressed the argument naming convention in a new commit.

bin/scan-return-of-results/generate
kairstenfay: Addressed this in a new commit.

bin/scan-return-of-results/generate
kairstenfay: Good to know which paths to avoid, and I'm always happy to learn about another useful bash command!

Addressed in a new commit.

kairstenfay:
> LGTM. I pushed a few commits addressing the awkward `/cwd` hack. Take a look?

Thanks for tackling the gnarly Docker bits! I'll test once more.

kairstenfay:
Just tested locally with the Docker changes and it works great.
There are still some nice-to-haves that would improve development, but they're not as high priority as actually making PDFs faster, so I'm gonna merge now.

PR #43, opened 2020-05-01
SCAN Amazon addresses: export report and upload to AWS

joverlee521:
Exports the "Amazon Care Outgoing" Report from the English SCAN REDCap
project and uploads to AWS S3 bucket.

Currently only exports from the English project but identical reports
will be set up for the other languages in the future.


joverlee521:
[Generate test csv of addresses to share w/ Amazon](https://trello.com/c/5qPxGQgN/560-generate-test-csv-of-addresses-to-share-w-amazon)

joverlee521:
I was able to upload locally after following the [instructions](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html) for configuring a role in the config file. It doesn't look like the ARN can be provided through an environment variable, so do we just add this as a new profile to `~/.aws/config` on backoffice?

bin/scan-amazon-addresses/export-redcap-report
```diff
@@ -0,0 +1,22 @@
...
+: "${REDCAP_REPORT_ID:?The REDCAP_REPORT_ID environment variable is required.}"
+
+
+curl -fsSL -H "Content-Type: application/x-www-form-urlencoded" \
+      -H "Accept: application/json" \
+      -X POST \
+      -d @<(echo "token=$REDCAP_API_TOKEN") \
+      -d format=csv \
+      -d content=report \
+      -d @<(echo "report_id=$REDCAP_REPORT_ID") \
```
kairstenfay, line 19: I wonder if we want to pass the report ID as a parameter rather than an environment variable. 
I expect we don't want to hard code it because the ID will change between REDCap projects.

kairstenfay: Another note: the `REDCAP_REPORT_ID` isn't sensitive, so it would be good to commit to this repo.

joverlee521: I made it an environment variable so that it can get passed via `envdir` without having to provide extra arguments. 

I can add `REDCAP_REPORT_ID_EN` to this repo within `id3c-production/env.d/redcap-scan`

bin/scan-amazon-addresses/generate-amazon-addresses
```diff
@@ -0,0 +1,47 @@
...
+#
+# Exports the Amazon Care Outgoing Report from SCAN REDCap project and uploads
+# to Amazon S3 bucket.
+#
+set -euo pipefail
+
+: "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
+
+bin="$(dirname "$0")"
+addresses="${1:-s3://698378617677-prod-or-uwscan-intake/intake/sample.csv}"
```
tsibley, line 12: Since a `main()` function exists that's called with all command-line arguments, I'd expect `addresses` to be local to `main` and evaluated after the `--help` processing.

bin/scan-amazon-addresses/generate-amazon-addresses
```diff
@@ -0,0 +1,47 @@
...
+main() {
+    for arg; do
+        case "$arg" in
+            --help)
+                print-help
+                exit 0;;
+        esac
+    done
+
+    export-redcap-data | aws s3 cp - "$addresses" --acl bucket-owner-full-control
```
tsibley, line 23: Add a `--quiet` here to reduce log noise?

bin/scan-amazon-addresses/generate-amazon-addresses
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# usage: generate-amazon-addresses
```
tsibley, line 2: Usage line could include `[<s3 output url>]` to indicate support for that.

crontabs/scan-amazon-addresses
```diff
@@ -0,0 +1,11 @@
...
+PATH=/usr/local/bin:/usr/bin:/bin
+
+LOG_LEVEL=warning
+
+# Base of our env.d directories
+ENVD=/opt/backoffice/id3c-production/env.d
+
+
+# Generate and upload CSV of SCAN addresses for outgoing Amazon Care
+0 6 * * * ubuntu chronic envdir $ENVD/redcap-scan/ envdir $ENVD/scan-aws-role/ /opt/backoffice/bin/scan-amazon-addresses/generate-amazon-addresses
```
tsibley, line 11: `scan-aws-role` seems not that specific. Could we rename to `amazon-care-fulfillment` or something similar?

id3c-production/env.d/scan-aws-role/AWS_PROFILE
```diff
@@ -0,0 +1 @@
+scan
```
tsibley, line 1: This profile name is also not very specific.

PR #46, opened 2020-05-07
So ko ru

kairstenfay:
Make backoffice edits described by https://github.com/seattleflu/documentation/blob/master/scan.md to prepare for three new languages.


kairstenfay:
Rebased onto master and bumped the Docker image version again as a result.

PR #49, opened 2020-06-09
Scan switchboard

kairstenfay:
Deploy the SCAN Switchboard app (a uvicorn/ASGI app) to the backoffice server using systemd. uWSGI unfortunately does not support ASGI.


crontabs/scan-switchboard
```diff
@@ -0,0 +1,12 @@
...
+
+# Point Pipenv at the SCAN Switchboard environment
+PIPENV_PIPFILE=/opt/scan-switchboard/Pipfile
+
+# SCAN Switchboard uses the same environment variables as id3c-production
+ENVD=/opt/backoffice/id3c-production/env.d
+
+
+# Refresh the SCAN Switchboard SQLite database every minute
+* * * * * ubuntu chronic envdir $ENVD/redcap-scan/ (cd /opt/scan-switchboard/; make -B)
```
kairstenfay, line 12: I couldn't figure out how to run the Makefile using an absolute path. I think this will work but I'm not sure it's ideal.

kairstenfay:
[Switchboard: Deploy to backoffice](https://trello.com/c/m38id0dV/581-switchboard-deploy-to-backoffice)

crontabs/scan-switchboard
tsibley: Use `make -BC /opt/scan-switchboard`. The `-C` option changes to the given directory first before looking for the Makefile or running any recipes. (This option is somewhat of a convention for tools particularly sensitive to the working directory. For example, git also has a `-C` option which does the same thing.)

kairstenfay: Cool, I did not know about that convention.

kairstenfay:
Fixed up requested changes into the latest commit.

PR #50, opened 2020-06-10
scan-return-of-results: Export only SCAN IRB data

kairstenfay:
Stop exporting RoR data for the REDCap SCAN project (currently shut down
by the FDA) and instead export from the new SCAN IRB (a.k.a. "research"
or "study") REDCap project. Currently, English is the only language
in production for this new study.

These changes also co-occur with a new lab result report template for
the SCAN IRB project. Bump the Docker image version used in
`generate-pdfs` accordingly.

Add a new env.d directory for the environment variables required for
pulling from all REDCap SCAN IRB projects.


joverlee521:
> > I think I have a slight preference for separating redcap-scan and redcap-scan-irb in favor of shorter file names, but I'm open here.
> 
> > Additionally, I think whichever env.d structure we opt for (all in redcap-scan or a split into redcap-scan-irb) will support pulling data from both studies.
> 
> ~Hmm, even if we have them in separate directories, the variable name would still have to be different in order for them to be used in the same command.~
> 
> Nvm, they shouldn't be in the same command since they would have to use different report templates. 🤦‍♀️

Ah, looking at the new PR for the switchboard (https://github.com/seattleflu/scan-switchboard/pull/6), it seems like the environment variable names need to be different regardless of whether they live in separate directories within `env.d`

kairstenfay:
> Ah, looking at the new PR for the switchboard ([seattleflu/scan-switchboard#6](https://github.com/seattleflu/scan-switchboard/pull/6)), it seems like the environment variable names need to be different regardless of whether they live in separate directories within `env.d`

Yes, you're right. I'll change my end.

kairstenfay:
> Yes, you're right. I'll change my end.

Updated in the latest commit.

kairstenfay:
> The `env.d/redcap-scan-irb` directory added by the first commit should be removed by the second too.

Fixed!

kairstenfay:
[Updates to Securelink results portal](https://trello.com/c/IIlAtrAa/608-updates-to-securelink-results-portal)

PR #52, opened 2020-06-16
Turn off the website-science-map cronjob

kairstenfay:
The SFS website has not been active for some time. Stop installing the
website-science-map cronjob, but keep the file in case we want to turn
the job back on again.


tsibley:
Note that the crontab will also have to be manually removed from its installed location when deploying this change.

I would probably delete the crontab and associated export script from the repo. We can always pull them out of git history later.

kairstenfay:
> Note that the crontab will also have to be manually removed from its installed location when deploying this change.

Noted 

> I would probably delete the crontab and associated export script from the repo. We can always pull them out of git history later.

The crontab and export script files are deleted in the latest version of the commit.

tsibley:
Oh, would you also delete `s3://seattle-flu-study/samples.geojson` when you deploy this?

kairstenfay:
> Oh, would you also delete `s3://seattle-flu-study/samples.geojson` when you deploy this?

The asociated crontab file on the backoffice server and s3 geojson file have both been deleted.

PR #55, opened 2020-06-30
Notification fatigue

tsibley:
See commit messages. I think this should stop floods of cron errors.


tsibley:
Happy to explain at any time!

Re-pushed to update `fatigue` to use a lockfile so that state isn't trashed if multiple copies run at the same time for the same command. (Shouldn't normally happen, but could if a job starts taking a while and the next occurrence starts before the previous is finished.)

PR #56, opened 2020-07-02
scan-return-of-results: Return Vietnamese results

kairstenfay:
Return results from the Vietnamese SCAN IRB REDCap project.

Depends on https://github.com/seattleflu/lab-result-reports/pull/19


kairstenfay:
[SCAN IRB: Vietnamese](https://trello.com/c/yFg6Iyed/648-scan-irb-vietnamese)

kairstenfay:
The master branch is already using the latest Docker tag which includes a Vietnamese template, so there's no need to bump the image tag here.

PR #57, opened 2020-07-06
Scan irb/ru

kairstenfay:
Export results from the SCAN IRB Russian project. 

~Based off of #56.~
Depends on https://github.com/seattleflu/lab-result-reports/pull/20


kairstenfay:
[SCAN IRB: Russian](https://trello.com/c/gKFKht0Q/650-scan-irb-russian)

bin/scan-return-of-results/generate-pdfs
```diff
@@ -50,7 +50,7 @@ main() {
             --env=LOG_LEVEL \
             --user "$(id -u):$(id -g)" \
             "${mounts[@]}" \
-            seattleflu/lab-result-reports:build-21 \
+            seattleflu/lab-result-reports:build-23 \
```
joverlee521, line 53: Are we skipping `build-22`?

kairstenfay: Build 22 is in #56. I may or may not create two new build tags depending on when the external reviewers approve the new template.

kairstenfay:
I rebased onto master instead of onto the Vietnamese project, because we stopped doing translations for the time being. Russian is an exception.

PR #58, opened 2020-07-11
Group enroll redcap dets

kairstenfay:
Create a cron job that uploads "fake" REDCap DETs for SCAN IRB group enrollment data. 
Schedule it for once an hour. 

Depends on https://github.com/seattleflu/id3c/pull/156

### DIFF
--- a/bin/generate-and-upload-group-enrollment-redcap-dets
+++ b/bin/generate-and-upload-group-enrollment-redcap-dets
@@ -1,0 +1,55 @@
+#!/bin/bash
+# usage: ./bin/generate-and-upload-group-enrollment-redcap-dets
+#
+# Generates group enrollment (an event type) REDCap DETs from the SCAN IRB
+# REDCap projects for records that have been created and/or modified since
+# 25 hours ago and uploads them to ID3C.
+#
+
+set -euo pipefail
+
+# Check this early before we do any work rather than failing at the end.
+: "${REDCAP_API_URL:?The environment variable REDCAP_API_URL is required.}"
+
+# Absolute path to our containing repo
+base="$(cd "$(dirname "$0")/.."; pwd)"
+
+
+main() {
+    for arg; do
+        case "$arg" in
+            -h|--help)
+                print-help
+                exit 0;;
+        esac
+    done
+    upload-to-id3c <(generate-redcap-dets)
+}
+
+generate-redcap-dets() {
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_EN" id3c redcap-det generate --project-id 22461 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_ES" id3c redcap-det generate --project-id 22475 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_ZH_HANT" id3c redcap-det generate --project-id 22474 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
+}
+
+upload-to-id3c() {
+    id3c redcap-det upload "$1"
+}
+
+print-help() {
+    # Print the help comments at the top of this file ($0)
+    local line
+    while read -r line; do
+        if [[ $line =~ ^#! ]]; then
+            continue
+        elif [[ $line =~ ^# ]]; then
+            line="${line/##/}"
+            line="${line/# /}"
+            echo "$line"
+        else
+            break
+        fi
+    done < "$0"
+}
+
+main "$@"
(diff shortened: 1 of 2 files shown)


bin/generate-and-upload-group-enrollment-redcap-dets
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# usage: ./bin/generate-and-upload-group-enrollment-redcap-dets
+#
+# Generates group enrollment (an event type) REDCap DETs from the SCAN IRB
+# REDCap projects for records that have been created and/or modified since
+# 25 hours ago and uploads them to ID3C.
```
kairstenfay, line 6: I didn't actually ask why you all chose 25 hours for the --since date for these DETs. Was it the gap between scheduled runs + 1 hour?

kairstenfay:
[Set up DET generation for group enrollments](https://trello.com/c/m4HM18kU/660-set-up-det-generation-for-group-enrollments)

bin/generate-and-upload-group-enrollment-redcap-dets
joverlee521: Ah, yes we chose 25 hours for the UW retros because it was running every 24 hours. The extra 1 hour is just extra insurance that we don't miss any records.

Since we are running this script once an hour, we can decrease the pull to since 2 hours ago?

bin/generate-and-upload-group-enrollment-redcap-dets
```diff
@@ -0,0 +1,55 @@
...
+#
+
+set -euo pipefail
+
+# Check this early before we do any work rather than failing at the end.
+: "${REDCAP_API_URL:?The environment variable REDCAP_API_URL is required.}"
```
tsibley, line 12: The GitHub diff doesn't show that this file was copied with changes from `generate-and-upload-uw-retro-redcap-dets`, but a local diff does:

image

In the source file, the check was intentionally after the `--help` processing so that one could run `generate-and-upload-group-enrollment-redcap-dets --help` without defining the environment variable first.

bin/generate-and-upload-group-enrollment-redcap-dets
```diff
@@ -0,0 +1,55 @@
...
+                exit 0;;
+        esac
+    done
+    upload-to-id3c <(generate-redcap-dets)
+}
+
+generate-redcap-dets() {
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_EN" id3c redcap-det generate --project-id 22461 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_ES" id3c redcap-det generate --project-id 22475 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
+    REDCAP_API_TOKEN="$REDCAP_API_TOKEN_IRB_ZH_HANT" id3c redcap-det generate --project-id 22474 --event "group_enroll_arm_4" --since-date "`date +"%F %T" -d "25 hours ago"`" --include-incomplete
```
tsibley, line 32: To be a little bit safer, I'd compute the "25 hours ago" timestamp once before making any of the `id3c redcap-det generate` invocations so that they all use the same `--since-date` timestamp instead of slightly drifting. The drift probably won't be consequential, but if one of the first generations takes a long time for some pathological reason, then subsequent invocations could use a timestamp that misses records.

crontabs/id3c-production
```diff
@@ -48,6 +40,9 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 # Create and upload REDCap DETs for UW retrospectives once a day at 4 A.M.
 0 4 * * * ubuntu pipenv run chronic envdir $ENVD/redcap-uw-retrospectives/ generate-and-upload-uw-retro-redcap-dets
 
+# Create and upload REDCap DETs for SCAN IRB group enrollments once an hour at a quarter past.
+15 */1 * * * ubuntu pipenv run chronic envdir $ENVD/redcap-scan/ generate-and-group-enrollment-redcap-dets
```
jccraft-uw, line 44: Just for my own learning, do we need "*/1" for the hour in there? 
Would "15 * * * *" run at the quarter after every hour?

joverlee521: @jccraft-uw Yes, both would work here. Here's a neat site to play around with cron expressions: https://crontab.guru/

joverlee521:
This cronjob is no longer needed because the `back_end_mail_scans` instrument is getting completed the same way as usual records. So we would received DETs when the `back_end_mail_scans` is complete, which is when we would ingest the REDCap record anyways! @kairstenfay I apologize for misleading you to do extra work here 🙏

crontabs/id3c-production
kairstenfay: Yep, I should have written `15 * * * *` instead to eliminate the redundancy of `*/1`*

kairstenfay:
No problem :)

bin/generate-and-upload-group-enrollment-redcap-dets
kairstenfay: Ah, that makes sense.

PR #60, opened 2020-07-16
Ingest specimen manifest sheet 7

kschwabefry:
Start ingesting aliquoting sheet 7. This includes new columns `swab_type` and `collection_matrix`.

https://trello.com/c/XjWs7Yji/677-update-specimen-manifest-ingest-to-put-swab-type-and-media-into-warehousesampledetails

### DIFF
--- a/etc/uw-retrospectives-manifest.yaml
+++ b/etc/uw-retrospectives-manifest.yaml
@@ -58,3 +58,13 @@ extra_columns:
   mrn: mrn
   accession_no: accession
   sample_origin: sample_origin
+---
+workbook: s3://fh-pi-bedford-t/seattleflu/bbi/2019_2020_sfs_aliquoting_7.xlsx
+sheet: aliquoting
+sample_column: sample_id
+extra_columns:
+  barcode: sample_id
+  collection_date: collection_date
+  mrn: mrn
+  accession_no: accession
+  sample_origin: sample_origin


kschwabefry:
[Update specimen manifest ingest to put swab type and media into warehouse.sample.details](https://trello.com/c/XjWs7Yji/677-update-specimen-manifest-ingest-to-put-swab-type-and-media-into-warehousesampledetails)

etc/uw-retrospectives-manifest.yaml
```diff
@@ -58,3 +58,15 @@ extra_columns:
...
+sheet: aliquoting
+sample_column: sample_id
+extra_columns:
+  barcode: sample_id
+  collection_date: collection_date
+  mrn: mrn
+  accession_no: accession
+  sample_origin: sample_origin
+  swab_type: swab_type
+  collection_matrix: collection_matrix
```
joverlee521, line 72: Ah, sorry I wasn't clear about this. These two new columns do not need to be added to the UW-retro manifest. Just the new workbook with the same columns as before. 

For more context, this YAML determines where/what we pull from UW-retro data to upload to the Clinical Data pulls REDCap project. We do not need the two new columns for this side.

kschwabefry: No worries, I'll can go ahead and drop those columns and update the commit message.

So the UW retrospective data in that redcap project is a more limited subset of what's in the lab's spreadsheet then? It looks like this manifest used by backoffice/bin/import-uw-retrospectives-to-redcap

PR #61, opened 2020-07-17
Import redcap user

kairstenfay:
See commits.


dev/import-redcap-user
```diff
@@ -0,0 +1,106 @@
...
+
+
+parser = argparse.ArgumentParser(
+    description= __doc__,
+    formatter_class=argparse.RawTextHelpFormatter
+)
+
+parser.add_argument("--unity-id",
+    metavar="<unity-id>",
+    help="A user's UW Unity ID")
```
tsibley, line 89: UW Unity ID? I'm not familiar with that identity system at UW. Do you mean UW NetID?

kairstenfay: Uh, I think so? I'm not sure where I got to thinking "Unity" in my head.

jccraft-uw: UW's main person ID is the "UW NetID"

kairstenfay: Thanks, guys! I'll fix it.

jccraft-uw:
I'm thinking about how a user (like Misja) would want to give RedCAP access. If she wants to give one user access to one project she can do that in RedCAP itself. What is painful to do in RedCAP are bulk operations like adding one user to multiple projects or adding multiple users to one or multiple projects. Without a tool this is a painfully manual and error prone process.

I'm thinking that we could have the user populate an Excel sheet where one row would be a single user's access to one or multiple projects. Then the user could copy/paste/drag to give the same access to multiple users. This would give the user a way to define the access the users should have. We would read this in and split operations out by project and by user ... for each project ... for each user. 

An advantage is that we would have a document from the user telling us which users should get which access. For compliance reasons, we would want to store these sheets so that later if someone questions why user X has access to project Y we would have a record to say who granted that access.

We could simplify this for the user by saying "all instruments" instead of requiring the user to list them out. We'd want to know how Misja typically grants access ... does she typically give all users the same access to all instruments, or does she configure instrument by instrument.

tsibley:
> For compliance reasons, we would want to store these sheets so that later if someone questions why user X has access to project Y we would have a record to say who granted that access.

FWIW, I believe the necessary compliance information is already tracked in REDCap itself.

tsibley:
> It's the RedCAP API token that determines which user gets recorded as having done the RedCAP API action, right?

Yep.

> Would this script get run by one of us or by the user wanting to grant access to other users? If I were to run this script for Misja I'd want it recorded somehow that Misja was the one who granted these users access, not me. I was just the middle person running the script.

Agreed. I think it'd be Misja running it, ideally.

dev/add-user-to-all-projects
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# usage:
+
+# Required environment variables:
+#   - REDCAP_API_URL
+#   - USER_PROJECT_PERMISSIONS: The UW NetID of the user whose permissions you
```
jccraft-uw, line 6: Is this description for the NET_ID env var instead of the USER_PROJECT_PERMISSIONS one?

kairstenfay: Good question. I couldn't think of a good name for this variable. Underneath the hood, it's the `NET_ID` var but the one that gets fed into the export script, not the import script.

dev/add-user-to-all-projects
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# usage:
```
kairstenfay, line 2: TODO

dev/add-user-to-all-projects
jccraft-uw: Gotcha!

jccraft-uw: How about something like "template_Net_ID" or something like that. This is the "template" or "example" user whose permissions are getting "cloned".

kairstenfay: I like that suggestion. It's semantically clearer what purpose the variable serves.

tsibley: Spitballing on names:

- `TEMPLATE_USER` and `NEW_USER`
- `EXISTING_USER` and `NEW_USER`
- `SRC_USER` and `DST_USER`

dev/export-redcap-user
```diff
@@ -0,0 +1,94 @@
...
+# usage: ./dev/export-redcap-user --net-id <net-id>
+#
+# Exports the given REDCap user's metadata and permissions from the project
+# specified by the REDCap API token environment variable.
+# Useful for deriving permissions configurations for ./dev/import-redcap-user.
+# The returned user's information is printed to stdout, so you'll likely want to
+# redirect to a new file.
+#
+# Raises a StopIteration error if the given user does not exist in the REDCap
+# project.
```
tsibley, line 11: This is outdated, I think. The code below raises an Exception (which is better than letting Python's internal StopIteration exception through).

dev/export-redcap-user
```diff
@@ -0,0 +1,94 @@
...
+
+# We only want permissions for the user with the given UW NetID
+try:
+    index = next(index for index, item in enumerate(users)
+        if item["username"] == f"{args.net_id}@washington.edu")
+except StopIteration as e:
+    raise Exception(f"No user found in the given REDCap project with --net-id {args.net_id}")
+
+permissions = users[index]
+print(json.dumps(permissions))
```
tsibley, line 94: FWIW, the use of `index` here is a bit unnecessary.

```diff
@@ -85,10 +85,8 @@ users = fetch()
 
 # We only want permissions for the user with the given UW NetID
 try:
-    index = next(index for index, item in enumerate(users)
-        if item["username"] == f"{args.net_id}@washington.edu")
+    user = next(u for u in users if u["username"] == f"{args.net_id}@washington.edu")
 except StopIteration as e:
     raise Exception(f"No user found in the given REDCap project with --net-id {args.net_id}")
 
-permissions = users[index]
-print(json.dumps(permissions))
+print(json.dumps(user))

```

dev/add-user-to-all-projects
kairstenfay: I like `TEMPLATE_USER` and `NEW_USER`. I'll update, thanks!

kairstenfay: Actually, "new" is a bit of a misnomer because the user can exist already in REDCap and have their permissions imported. So, SRC and DST it is!

dev/export-redcap-user
kairstenfay: Ah yes I see that now. That's what I get for copying code from the internet :upside_down_face:. Thanks!

dev/export-redcap-user
kairstenfay: Yes, thank you

dev/add-user-to-all-projects
kairstenfay: Fixed up in the latest commit.

dev/export-redcap-user
kairstenfay: Fixed in the latest commits.

dev/export-redcap-user
kairstenfay: Fixed in the latest commits.

dev/import-redcap-user
```diff
@@ -0,0 +1,116 @@
...
+args = parser.parse_args()
+
+with open(args.permissions, 'r') as f:
+    permissions = json.loads(f.read())
+
+new_user_info = {
+    'username': f"{args.net_id}@washington.edu",
+    'email': args.email,
+    'firstname': args.firstname,
+    'lastname': args.lastname
```
joverlee521, line 112: I don't see `email`, `firstname`, and `lastname` in the list of all available attributes in the Import Users API docs. (It could just be the docs are incomplete 🤷) Do the extra fields not cause an error?

dev/export-redcap-user
```diff
@@ -0,0 +1,91 @@
...
+# Exports the given REDCap user's metadata and permissions from the project
+# specified by the REDCap API token environment variable.
+# Useful for deriving permissions configurations for ./dev/import-redcap-user.
+# The returned user's information is printed to stdout, so you'll likely want to
+# redirect to a new file.
+#
+# Raises an Exception if the given user does not exist in the REDCap project.
+#
+# See the REDCap Export Users API Documentation for more info:
+#   https://redcap.iths.org/redcap_v10.1.4/API/help.php?content=exp_users
```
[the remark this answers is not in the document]
kairstenfay: Good point. Do you think it's more helpful to leave it knowing it may one day break or exclude the links altogether? If we leave them, at least we'd have documented a specific API version that used to work.

dev/import-redcap-user
kairstenfay: Ah, good catch! I overlooked that bit. It turns out REDCap/ITHS automatically fills in the firstname, lastname, and email info if you import a user with their UW NetID. Those extra fields did not cause an error.

kairstenfay: I just re-pushed without firstname, lastname, and email. Much simpler :)

dev/export-redcap-user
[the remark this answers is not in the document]
joverlee521: Ah, that's true, it would be useful to know which version was used. Maybe exclude the link but document the version.

kairstenfay:
Just force-repushed a couple of small copy-edits. Done editing.

dev/export-redcap-user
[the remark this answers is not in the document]
kairstenfay: Done in latest commits.

PR #63, opened 2020-07-20
Run scan-switchboard make every 5 mins

jccraft-uw:


crontabs/scan-switchboard
```diff
@@ -12,4 +12,4 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 
 
 # Refresh the SCAN Switchboard SQLite database every minute
```
kairstenfay, line 14: this comment needs to be updated

jccraft-uw: Fixed. Thanks for calling that out.

PR #65, opened 2020-07-21
Return of results processing for SCAN kiosk/in-person enrollments

tsibley:
See commit messages for details.

I believe the barcode handling is correct and appropriately cautious, but I'm not super happy with the Pandas bits. I wonder if there's a clearer approach. Happy to hear suggestions!

This is ready for review, although there's one remaining commit/todo item before this merges:

- [x] Add `swab_type` and `staff_collected` fields (or whatever their final names are) to the output of `transform`. Depends on https://github.com/seattleflu/id3c-customizations/pull/119.

### DIFF
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -10,8 +10,9 @@ from typing import NamedTuple, Optional
 
 
 REDCAP_API_URL = environ["REDCAP_API_URL"]
-REDCAP_API_TOKEN = lambda lang: environ[(
-    ("REDCAP_API_TOKEN_IRB_%s" % lang)
+REDCAP_API_TOKEN = lambda lang, collection_via: environ[((
+    "REDCAP_API_TOKEN_IRB_%s" % lang if collection_via == "mail" else
+    "REDCAP_API_TOKEN_IRB_%s_%s" % (collection_via, lang))
         .upper()
         .replace("-", "_")
 )]
@@ -23,6 +24,7 @@ FIELDS = [
     "birthday",
     "utm_tube_barcode_2",
     "reenter_barcode",
+    "reenter_barcode_2",
     "return_utm_barcode",
     "contacted",
 ]
@@ -36,18 +38,23 @@ REQUIRED_FIELDS = [
 
 class ScanProject(redcap.Project):
     lang: str
+    collection_via: str
 
-    def __init__(self, project_id: str, lang: str) -> None:
-        super().__init__(REDCAP_API_URL, REDCAP_API_TOKEN(lang), project_id)
+    def __init__(self, project_id: str, lang: str, collection_via: str) -> None:
+        super().__init__(REDCAP_API_URL, REDCAP_API_TOKEN(lang, collection_via), project_id)
         self.lang = lang
+        self.collection_via = collection_via
 
 
 def main():
     projects = [
         # SCAN (research study)
-        ScanProject(22461, "en"),
-        ScanProject(22475, "es"),
-        ScanProject(22474, "zh-Hant"),
+        ScanProject(22461, "en", "mail"),
+        ScanProject(22475, "es", "mail"),
+        ScanProject(22474, "zh-Hant", "mail"),
+
+        # SCAN kiosks
+        ScanProject(23089, "en", "kiosk"),
     ]
 
     for project_records in map(fetch_records, projects):
@@ -60,7 +67,10 @@ def fetch_records(project):
         if not all(len(record[field]) for field in REQUIRED_FIELDS):
             continue
 
-        yield record
+        yield {
+            "project_collection_via": project.collection_via,
+            **record,
+        }
 
 
 if __name__ == "__main__":
--- a/bin/scan-return-of-results/export-redcap-scan
+++ b/bin/scan-return-of-results/export-redcap-scan
@@ -1,7 +1,7 @@
 #!/bin/bash
 # usage: export-redcap-scan
 #
-# Exports participant information from the SCAN REDCap projects as CSV.
+# Exports participant information from the SCAN REDCap projects as NDJSON.
 #
 set -euo pipefail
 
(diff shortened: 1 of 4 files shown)


tsibley:
[Update return of results PII join to include new Greek system REDCap project](https://trello.com/c/cT3SR8cs/681-update-return-of-results-pii-join-to-include-new-greek-system-redcap-project)

bin/scan-return-of-results/export-redcap-scan
```diff
@@ -1,25 +1,77 @@
...
-      "$REDCAP_API_URL"
+import json
+from id3c.cli import redcap
+from os import environ
+from typing import NamedTuple, Optional
+
+
+REDCAP_API_URL = environ["REDCAP_API_URL"]
+REDCAP_API_TOKEN = lambda lang, collection_via: environ[((
+    "REDCAP_API_TOKEN_IRB_%s" % lang if collection_via == "mail" else
```
jccraft-uw, line 14: This is the kind of metadata that I want us to look some day to centralize. Rather than have to build logic to determine the env var name it could be part of the definition of a RedCAP project.

bin/scan-return-of-results/export-redcap-scan
```diff
@@ -1,25 +1,77 @@
...
-# Exports participant information from the SCAN REDCap projects as CSV.
+# Exports participant information from the SCAN REDCap projects as NDJSON.
 #
-set -euo pipefail
-
-: "${REDCAP_API_URL:?The REDCAP_API_URL environment variable is required.}"
-: "${REDCAP_API_TOKEN:?The REDCAP_API_TOKEN environment variable is required.}"
-
-fields="record_id,participant_first_name,participant_last_name,birthday,pre_scan_barcode,utm_tube_barcode_2,reenter_barcode,return_utm_barcode,contacted"
-filters="[participant_first_name] != '' and [participant_last_name] != '' and [birthday] != ''"
```
jccraft-uw, line 12: Just confirming that it's expected that we're not using 'filters' now that we're using id3c.cli.recap.

bin/scan-return-of-results/transform
```diff
@@ -14,18 +14,73 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+        "return_utm_barcode",
+        "utm_tube_barcode_2",
+        "reenter_barcode",
+        "reenter_barcode_2"}
+
+    for barcode_field in barcode_fields:
+        if barcode_field in redcap_data:
+            redcap_data[barcode_field] = normalize_barcode(redcap_data[barcode_field])
+        else:
+            redcap_data[barcode_field] = pd.Series(dtype = "string")
```
jccraft-uw, line 41: Just confirming: a given barcode field can contain an array and not just a string, thus pd.Series here?

kairstenfay: `pd.Series` is like a column in Pandas

jccraft-uw: This thread helped me understanding why we might not want to just do redcap_data[barcode_field] = np.nan

https://stackoverflow.com/questions/16327055/how-to-add-an-empty-column-to-a-dataframe

"If you have an empty dataframe, this solution makes sure that no new row containing only NaN is added."

kairstenfay: Yep, we try to stick with Pandas dtypes rather than introducing numpy ones.

bin/scan-return-of-results/transform
```diff
@@ -15,7 +15,8 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
     prepared in the specifications of UW Lab Med's return of results portal.
     """
     redcap_data = (
-        pd.read_csv(redcap_file, dtype = 'string')
+        pd.read_json(redcap_file, lines = True, dtype = False, convert_dates = False)
+        .astype("string")
```
kairstenfay, line 19: I feel like we've probably hashed this out before, but could you please remind me why you cast to "string" after reading in the data rather than providing `dtype = "string"` on read?

bin/scan-return-of-results/export-redcap-scan
tsibley: Ah, correct. That filtering now happens client-side instead of server-side. Code is: https://github.com/seattleflu/backoffice/pull/65/commits/d705c1c8c3c1bd1bad7316208c020e4d7f68f334#diff-59ff1b7e08db1e5b3948ca4a0489173bR60-R61

This seemed more expedient than adding support for `filters` to our `id3c.cli.redcap` package and bumping deps, etc.

bin/scan-return-of-results/transform
tsibley: I wish I could, but Pandas doesn't provide a consistent API here! The support for initial dtype-ing during read depends on the `read_*` function. For `read_json`, dtype is documented as:

```     
dtype
    bool or dict, default None

    If True, infer dtypes; if a dict of column to dtype, then use those;
    if False, then don’t infer dtypes at all, applies only to the data.

    For all orient values except 'table', default is True.
```

where as `read_csv` documents it as:

```
dtype
    Type name or dict of column -> type, optional

    Data type for data or columns.
    E.g. {‘a’: np.float64, ‘b’: np.int32, ‘c’: ‘Int64’}

    Use str or object together with suitable na_values settings to preserve
    and not interpret dtype. If converters are specified, they will be applied
    INSTEAD of dtype conversion.
```

The difference does make sense when you consider that different file formats have different inherent typing available. CSV is all strings, while JSON has dictionaries, lists, numbers, strings, booleans, and null.

So `read_json(…, dtype = False, convert_dates = False)` says "don't try to do anything fancy with dtype on read". For our data, we get a dataframe of mostly the `object` dtype, which are then converted to strings with `.astype("string")`.

tsibley: > I feel like we've probably hashed this out before, but could you please remind me…

Oh, and don't feel like you should have remembered! I spent a good bit of time reading the docs closely and testing Pandas behaviour to boil it down to those lines of code and get it to stop messing with values in ways that weren't expected.

kairstenfay: Thank you for this explanation!

tsibley:
I've added one final commit to include `swab_type` and `staff_collected`.

tsibley:
One final item:

- [ ] Bump version of lab-result-reports image once {available, known, https://github.com/seattleflu/lab-result-reports/pull/23 is merged}.

PR #66, opened 2020-07-24
Export Seattle neighborhoods as svgs

kairstenfay:
Add two scripts to export the Seattle neighborhood shapefiles in ID3C as
flat SVG files to be used by Formative in the public-facing dashboard.

=======================================

These SVGs look ok in Inkscape but are appear rather small on import regardless of the width and height I assign in the viewBox. This should be fine considering that SVGs are by definition scalable. Any thoughts here?

### DIFF
--- a/bin/seattle-neighborhood-svgs/create-svgs
+++ b/bin/seattle-neighborhood-svgs/create-svgs
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+"""
+Create SVG elements from exported ID3C paths.
+"""
+import os
+import argparse
+import pandas as pd
+from pathlib import Path
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description= __doc__,
+        formatter_class=argparse.RawTextHelpFormatter
+    )
+
+    parser.add_argument("input",
+        metavar="<input.csv>",
+        help="A CSV export of the shipping.seattle_neighborhood_districts_v1 "
+            "view.")
+    parser.add_argument("--output-dir",
+        metavar="<output-directory>",
+        default=os.getcwd(),
+        required=False,
+        help="An optional directory under which to save the created SVG files. "
+            "By default, saves to the current working directory.")
+
+    args = parser.parse_args()
+
+    neighborhoods = pd.read_csv(args.input)
+
+    neighborhoods['svg'] = neighborhoods.apply(
+        lambda row: f'<svg viewBox="0 0 100 100"><path d="{row.svg_path}" /></svg>', axis='columns'
+    )
+
+    for _, row in neighborhoods.iterrows():
+        with open(f"{Path(args.output_dir) / row.identifier}.svg", 'w') as f:
+            f.write(row.svg)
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Formative data request](https://trello.com/c/57EgipiF/658-formative-data-request)

bin/seattle-neighborhood-svgs/export-id3c-svg-paths
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
+    \copy (select * from shipping.seattle_neighborhood_districts_v1) to pstdout with (format csv, header);
```
kairstenfay, line 4: I don't _think_ I need to use the centroid column, but I'll wait to drop it from this export until I confirm that.

kairstenfay:
Ah, I think I finally understand what @tsibley meant when he said to transform the coordinate system of the paths into document space. Would this require using the centroid?

bin/seattle-neighborhood-svgs/create-svgs
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+"""
+Create SVG elements from exported ID3C paths.
+"""
+import os
+import argparse
```
jccraft-uw, line 6: argparse versus Click. Does the team have any rules about which to use?

kairstenfay: Good question! My assumption (although I could be mistaken) is that argparse is lighter weight and more appropriate for solo scripts like this one, whereas Click is better for forming a more comprehensive CLI tool (hence, why we use it in id3c and id3c-customizations).

jccraft-uw: I agree with you about argparse being a lot lighter weight.

tsibley: I'd echo what @kairstenfay said. Part of what makes argparse lighter weight is that it's in the standard library, so no extra deps are needed.

jccraft-uw: Perfect! I didn't know if we had team norms about Click versus something lighter.

tsibley:
Thanks for looking into this, @kairstenfay. I think there might have been something lost in communication. Sorry about that; not sure where/when it got garbled. I believe Formative asked for a single SVG with all the shapes, not one per file. It also isn't something we need to make more than once, so we don't really need code that can generate it repeatedly. Copying/pasting the path syntax would be sufficient.

> These SVGs look ok in Inkscape but are appear rather small on import regardless of the width and height I assign in the viewBox. This should be fine considering that SVGs are by definition scalable. Any thoughts here?
>
> Ah, I think I finally understand what @tsibley meant when he said to transform the coordinate system of the paths into document space. Would this require using the centroid?

By adjusting the first two viewBox values (min x, min y) as well as the last two (height, width), you can adjust the document viewport to make the paths take up the whole viewport instead of a small portion of it. The centroid isn't required.

kairstenfay:
Closing in favor of https://github.com/seattleflu/id3c-customizations/pull/122

PR #67, opened 2020-07-28
Suppress no id warning for experimental barcodes

kschwabefry:
Filter out warnings for sample barcodes with suffix "*_exp", which is used by the lab for
experimental barcodes. These warnings are ignorable because these barcodes are not meant
to be ingested by the presence/absence and manifest ETL jobs.

https://trello.com/c/lolX0N3y/691-suppress-warning-for-barcodes-with-exp-suffix

### DIFF
--- a/id3c-production/logging.yaml
+++ b/id3c-production/logging.yaml
@@ -34,20 +34,20 @@ loggers:
       - INFO
 
 filters:
-  unknown sample warnings for controls from id3c etl presence-absence:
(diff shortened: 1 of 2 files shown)


kschwabefry:
[Suppress warning for barcodes with _exp suffix](https://trello.com/c/lolX0N3y/691-suppress-warning-for-barcodes-with-exp-suffix)

PR #70, opened 2020-08-06
export-id3c-scan-retun-results: exclude results with null swab_type

joverlee521:
Exclude results that are not pending and have a null swab_type.
This is done to prevent the return of results from being blocked by the
lack of swab type for never tested samples. The timing of a sample
getting marked as `never-tested` in REDCap and added to the
incident report is out of our control. This may lead to a block of
results since we cannot generate a PDF report without a known swab type.

A separate query is set up on Metabase to check hourly for null swab
types so that we know when we need to manually intervene.

### DIFF
--- a/bin/scan-return-of-results/export-id3c-scan-return-results
+++ b/bin/scan-return-of-results/export-id3c-scan-return-results
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -euo pipefail
 psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
-    \copy (select * from shipping.scan_return_results_v1) to pstdout with (format csv, header);
+    \copy (select * from shipping.scan_return_results_v1 where swab_type is not null or status_code = 'pending') to pstdout with (format csv, header);
 "


bin/scan-return-of-results/export-id3c-scan-return-results
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -euo pipefail
 psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
-    \copy (select * from shipping.scan_return_results_v1) to pstdout with (format csv, header);
+    \copy (select * from shipping.scan_return_results_v1 where swab_type is not null or status_code = 'pending') to pstdout with (format csv, header);
```
kschwabefry, line 4: Interesting. How do pending results get handled in the pipeline? Do we generate results for them?

joverlee521: Pending results are only displayed as "pending" on the portal website but do not get a PDF. Therefore we don't need to know swab type for them.

kschwabefry: Aha, that makes sense why they wouldn't need swab type. Thanks!

kairstenfay: This makes me wonder, should we set samples where swab_type is null to `pending` here for the purpose of the RoR portal?

joverlee521: @kairstenfay OH GOOD POINT. If we use this query, some people would get directed to the Error page since their barcode doesn't "exist" in the results csv.

joverlee521:
Closing this because it would result in participants directed to the Error page in the portal if their sample fall in the null swab type category.

PR #73, opened 2020-08-19
crontabs: Refresh materialized view shipping.scan_encounters_v1

joverlee521:
Should be deployed with https://github.com/seattleflu/id3c-customizations/pull/129

### DIFF
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -107,3 +107,4 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 # Run the refresh materialized view routine every 5 minutes include latest
 # ingested FHIR documents in the view
 */5 * * * * ubuntu fatigue --quiet pipenv run id3c refresh-materialized-view shipping fhir_questionnaire_responses_v1 --commit
+*/5 * * * * ubuntu fatigue --quiet pipenv run id3c refresh-materialized-view shipping scan_encounters_v1 --commit


joverlee521:
[Formative dashboard changes](https://trello.com/c/J7E4wxud/709-formative-dashboard-changes)

crontabs/id3c-production
```diff
@@ -107,3 +107,4 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 # Run the refresh materialized view routine every 5 minutes include latest
 # ingested FHIR documents in the view
 */5 * * * * ubuntu fatigue --quiet pipenv run id3c refresh-materialized-view shipping fhir_questionnaire_responses_v1 --commit
+*/5 * * * * ubuntu fatigue --quiet pipenv run id3c refresh-materialized-view shipping scan_encounters_v1 --commit
```
kairstenfay, line 110: This may be a dumb question, but should we be worried about both of these jobs starting at the same time? This concern also generally applies to tracking how many jobs start on the 0s or 5s. Would it be wiser to offset this new cronjob by one minute from the previous one? Or does this not really matter?

joverlee521: Hmm good question. `scan_encounters_v1` does depend on `fhir_questionnaire_responses_v1`, so I'm not sure if refreshing both at the same time would cause problems.

joverlee521: In discussion with @tsibley, there shouldn't be any major problems. If one refresh process locks out the other one, then the commands will be serialized.

PR #81, opened 2020-09-18
UW Re-opening ROR

joverlee521:
Add UW Reopening REDcap project to return of results pipeline.

### DIFF
--- a/bin/return-of-results/export-redcap-uw-reopening
+++ b/bin/return-of-results/export-redcap-uw-reopening
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+# usage: export-redcap-uw-reopening
+#
+# Exports participant information from the UW-Reopening REDCap project
+# (Husky Coronavirus Testing) as NDJSON.
+#
+import sys
+import json
+from id3c.cli.redcap import Project
+from os import environ
+from typing import Dict
+
+
+REDCAP_API_URL = environ["REDCAP_API_URL"]
+REDCAP_API_TOKEN = environ["REDCAP_API_TOKEN_UW_REOPENING"]
+REDCAP_PROJECT_ID = 23854
+
+ENROLLMENT_EVENT = "enrollment_arm_1"
+ENROLLMENT_FIELDS = [
+    "core_participant_first_name",
+    "core_participant_last_name",
+    "core_birthdate",
+]
+
+ENCOUNTER_EVENT = "encounter_arm_1"
+ENCOUNTER_FIELDS = [
+    "collect_barcode_kiosk",
+    "return_utm_barcode",
+]
+
+
+
+def main():
+    project = Project(REDCAP_API_URL, REDCAP_API_TOKEN, REDCAP_PROJECT_ID)
+
+    enrollment_records = fetch_enrollment_records(project)
+
+    # Update encounter data with enrollment data based on `record_id`
+    # Skip the record if it does not have enrollment data
+    for record in fetch_encounter_records(project):
+        record_id = record["record_id"]
+        enrollment_data = enrollment_records.get(record_id)
+        if not enrollment_data:
+            continue
+
+        record.update(enrollment_data)
+        print(json.dumps(record, indent = None, separators = ",:"), flush = True)
+
+
+def fetch_enrollment_records(project) -> Dict[str, dict]:
+    """
+    Fetch all enrollment records from *project* from the ENROLLMENT_EVENT.
+    Returns a dictionary of record ids and their corresponding fields for
+    records that have all ENROLLMENT_FIELDS completed.
+
+    Only expects one enrollment event per record_id, will drop any record_id
+    that has multiple records.
+    """
+    redcap_records: Dict[str,dict] = {}
+    duplicate_record_ids = set()
+
+    fields = ["record_id"] + ENROLLMENT_FIELDS
+    for record in project.records(events = [ENROLLMENT_EVENT], fields = fields, raw = True):
+        if not all(len(record[field]) for field in ENROLLMENT_FIELDS):
+            continue
+
+        record_id = record.pop("record_id")
+
+        if record_id in redcap_records:
+            duplicate_record_ids.add(record_id)
+
+        redcap_records[record_id] = { field: record[field] for field in ENROLLMENT_FIELDS }
+
+    if duplicate_record_ids:
+        for record_id in duplicate_record_ids:
+            del redcap_records[record_id]
+
+        print(f"Dropped {len(duplicate_record_ids)} REDCap records with duplicate"
+              f" enrollment record ids: {duplicate_record_ids}", file = sys.stderr)
+
+    return redcap_records
+
+
+def fetch_encounter_records(project):
+    """
+    Fetch all encounter records that are from the ENCOUNTER_EVENT.
+    Only include the encounter record if at least one of the ENCOUNTER_FIELDS
+    is completed.
+    """
+    fields = ["record_id"] + ENCOUNTER_FIELDS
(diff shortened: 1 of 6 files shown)


joverlee521:
[Add UW-Reopening REDCap to return-of-results](https://trello.com/c/znThLJKX/797-add-uw-reopening-redcap-to-return-of-results)

tsibley (approved):

bin/return-of-results/transform
```diff
@@ -85,18 +85,55 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+        "collect_barcode_kiosk",
+        "return_utm_barcode"}
+
+    for barcode_field in barcode_fields:
+        redcap_data[barcode_field] = normalize_barcode(redcap_data[barcode_field])
+
+    kiosk_barcodes = redcap_data["collect_barcode_kiosk"]
+    mail_barcodes = redcap_data["return_utm_barcode"]
+
+    barcodes = mail_barcodes.where(kiosk_barcodes.isnull(), kiosk_barcodes)
```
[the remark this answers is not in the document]
joverlee521: Ah, that's definitely clearer. Thanks!

bin/return-of-results/transform
```diff
@@ -112,43 +149,62 @@ def trim_whitespace(df: pd.DataFrame) -> pd.DataFrame:
...
 
-    participant_contacted = scan_data['contacted'] == 'yes'
+    participant_contacted = results['contacted'] == 'yes'
+    is_scan_data = results['source'] == 'scan'
 
-    scan_data.loc[(is_positive | is_inconclusive) & ~(participant_contacted), 'status_code'] = 'pending'
+    results.loc[(is_positive | is_inconclusive) & ~(participant_contacted) & (is_scan_data), 'status_code'] = 'pending'
 
-    scan_data.loc[scan_data['swab_type'].isnull(), 'status_code'] = 'pending'
+    results.loc[results['swab_type'].isnull(), 'status_code'] = 'pending'
```
jccraft-uw, line 187: Again, thanks for doing this weeks ago!

jccraft-uw (approved):

bin/return-of-results/export-redcap-uw-reopening
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+# usage: export-redcap-uw-reopening
```
kairstenfay, line 2: Would it be confusing to name this UW Reopening when the REDCap project is called "Husky Coronavirus Testing"?

kairstenfay: Additionally, do we expect future SFS projects to diverge significantly from this project? Could we name this script `export-redcap-sfs` and be inline with the envdir name?

bin/return-of-results/transform
```diff
@@ -85,18 +85,55 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
 
     # There should be absolutely no duplicates, but REDCap can't enforce this,
     # so warn and drop any that we find.
-    dups = barcodes.loc[barcodes.duplicated(keep = False)].dropna()
+    drop_duplicate_barcodes(barcodes)
```
kairstenfay, line 88: It feels a little odd to me to have this be the only part of our `transform` that does an inplace operation rather than re-assigning the output to the variable.

kairstenfay: I understand this was how it was written in a previous commit, though.

bin/return-of-results/transform
```diff
@@ -85,18 +85,55 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+    redcap_data['qrcode'] = barcodes
+    redcap_data['source'] = 'scan'
 
-        dups_count = len(dups)
-        dups_unique = list(dups.unique())
-        print(f"Dropped {dups_count:,} REDCap records with duplicated barcodes: {dups_unique}", file = sys.stderr)
+    return redcap_data[['qrcode', 'pat_name', 'birth_date', 'contacted', 'source']]
+
+
+def parse_uw_reopening_redcap(redcap_file) -> pd.DataFrame:
```
kairstenfay, line 96: Ditto re: my comment about generalizing this for SFS and not just UW Reopening a.k.a. Husky CT

bin/return-of-results/transform
```diff
@@ -112,43 +149,62 @@ def trim_whitespace(df: pd.DataFrame) -> pd.DataFrame:
...
-    is_inconclusive = scan_data['status_code'] == 'inconclusive'
+    is_positive = results['status_code'] == 'positive'
+    is_inconclusive = results['status_code'] == 'inconclusive'
 
-    participant_contacted = scan_data['contacted'] == 'yes'
+    participant_contacted = results['contacted'] == 'yes'
+    is_scan_data = results['source'] == 'scan'
 
-    scan_data.loc[(is_positive | is_inconclusive) & ~(participant_contacted), 'status_code'] = 'pending'
+    results.loc[(is_positive | is_inconclusive) & ~(participant_contacted) & (is_scan_data), 'status_code'] = 'pending'
```
kairstenfay, line 185: I can never remember order of operations for things like this. 
Is `& ~(participant_contacted) & (is_scan_data)` treated as one conditional or two separate ones?
I usually add an additional set of parentheses in these situations for clarity.
e.g.
`& (~participant_contacted & is_scan_data)`

tangentially:
I haven't tested yet but I'm not sure that the parens around `is_scan_data` are don't anything here.

joverlee521: Ah, you're right, the parens aren't doing anything here. I can remove them if they are confusing.

bin/return-of-results/export-redcap-uw-reopening
joverlee521: I'll add a comment that this is specifically for the Husky Coronavirus Testing REDCap project. 

Also commenting here for posterity, we do not expect future SFS projects to be the same format because the UW-reopening project is a longitudinal project with repeating events.

bin/return-of-results/transform
joverlee521: Yeah, I just did a copy&paste of the drop dupes logic. I can update the function to return the deduped dataframe instead of dropping it in place.

bin/return-of-results/transform
joverlee521: Updated in latest commit

bin/return-of-results/transform
joverlee521: Updated in latest commit

bin/return-of-results/transform
joverlee521: I expect the barcode fields will be different for other SFS projects. If it turns out that this can be generalized for other SFS projects, we can rename it in the future.

bin/return-of-results/transform
kairstenfay: Thank you!

bin/return-of-results/transform
kairstenfay: Sorry I was questioning if the code as it is written is doing what we'd expect or if the order of operations means that `~(participant_contacted)` gets resolved before `& is_scan_data`.

I think it might be more clear to put them together within parenthese, as in `& (~participant_contacted & is_scan_data)`

kairstenfay: (I may not have gotten the logic right here, I realize)

joverlee521: OH, I didn't think the order of operations matters here. Does that make the `.loc` filter faster?

kairstenfay: I just finally tested and confirmed the order of operations works as expected here. And at the script level (not function level), no time seems to be saved in wrapping parens around the final two conditionals, so never mind!

kairstenfay (approved):

PR #88, opened 2020-09-29
uw-ehs-report: Populate symptomatic_when_tested

jccraft-uw:
Replacing a placeholder NA value with the actual computed value for symptomatic_when_tested.
The ETL does not ingest the questionnaire items from the two instruments where these fields are, so I'm pulling directly from REDCap.

### DIFF
--- a/bin/uw-ehs-report/transform
+++ b/bin/uw-ehs-report/transform
@@ -91,9 +91,16 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
     enrollments.dropna(subset = {'netid'}, inplace = True)
 
     encounters = redcap_data.query('redcap_event_name == "encounter_arm_1"')[[ \
-        'record_id', \
-        'barcode' \
+        'record_id',
+        'barcode',
+        'illness_kiosk',
+        'illness_swabsend'
         ]]
+
+    # Set symptomatic_when_tested on encounters.
+    # Use '1' and '0' values to be consistent with the other boolean values we send to REDCap.
+    encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
+    encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
 
     joined_data = enrollments.merge(encounters, how='inner', on='record_id')
 
@@ -148,9 +155,6 @@ if __name__ == '__main__':
 
     id3c_data = pd.read_csv(args.id3c_data, dtype = 'string')
     id3c_data['barcode'] = normalize_barcode(id3c_data['barcode'])
-
-    # Placeholder field until the data pull gets implemented in ID3C.
-    id3c_data['symptomatic_when_tested'] = pd.NA
 
     # When you export a record set from Postgres as CSV, boolean columns
     # get values of 't' and 'f'. The EH&S transfer REDCap project expects
--- a/bin/uw-ehs-report/transform
+++ b/bin/uw-ehs-report/transform
@@ -195,7 +195,7 @@ if __name__ == '__main__':
         'on_campus_frequency_description', 'able_to_work_or_study_from_home_code',
         'able_to_work_or_study_from_home_description','is_uw_greek_member', 'uw_greek_house',
         'greek_other', 'housing_type', 'number_house_members', 'lives_with_uw_students_or_employees',
-        'uw_dorm_name', 'lives_in_uw_apartment',
+        'uw_dorm_name', 'uw_dorm_room_number', 'lives_in_uw_apartment',
         'uw_apartment_name', 'study_tier', 'symptomatic_when_tested'
 
         ]].to_csv(args.output, index=False)
--- a/bin/uw-ehs-report/export-redcap-uw-reopening
+++ b/bin/uw-ehs-report/export-redcap-uw-reopening
@@ -43,7 +43,11 @@ FIELDS = [
 
     # barcodes
     "collect_barcode_kiosk",
-    "return_utm_barcode"
+    "return_utm_barcode",
+
+    # non PII fields that we don't ingest into ID3C
+    "illness_kiosk",
+    "illness_swabsend"
 ]
 
 


bin/uw-ehs-report/transform
```diff
@@ -91,10 +91,16 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
-        'barcode' \
+        'record_id',
+        'barcode',
+        'illness_kiosk',
+        'illness_swabsend'
         ]]
 
+    # Set symptomatic_when_tested on encounters
+    encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
+    encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
```
kairstenfay, line 102: Is there a reason why we want to provide a string value of 0/1 instead of a more human readable value like "True" or "yes"?

jccraft-uw: We're sending boolean values of "1" and "0" for all the other boolean fields. See line 158.

kairstenfay: Gotcha. Thanks for commenting that!

joverlee521: Ah, I second this. The `symptomatic_when_tested` is currently a free text field.

kairstenfay: Ah yes I just noticed too we're casting 't' and 'f' to (integer) 1/0, but in the lines I've highlighted here, we're putting in string text of '1' and '0'.

jccraft-uw: We end up creating a CSV file that gets imported into REDCap. Everything ends up being a string in there.
I'll check with @kschwabefry  about the type of the symptomatic_when_tested field in REDCap. I expected it to be like the other fields to represent booleans.

jccraft-uw: And I remember why I did '1' instead of 1. When I used 1 it got written as "1.0" to the CSV. I wanted it to be "1" to be consistent with how we send other boolean values.

kschwabefry: The type of the symptomatic_when_tested field in REDCap is free text like Jover mentioned but now that we know its intended to be boolean, I will submit the change to make it a redcap "yes/no" (so the transformed data set would need to be coded as 1/0 (instead of say, 'true'/'false').

This would be consistent with how we are representing other boolean-type values in the transfer project. For the EHS transfer redcap project as a whole, we took the approach of making it look and act like a redcap project - using data types like yes/no, radio, and dates and adding descriptive labels on the fields. Since the EHS team works with the data in Redcap, we are optimizing the transferred data for the Redcap format over what it looks like from a database point of view.

kschwabefry: Trying to figure out why the integer 1 is converting to 1.0 thing might be worthwhile - consistency in this transform script with the other id3c booleans as 1 instead of "1" would be a nice to have.

joverlee521:
[Add uw_dorm_room_number back into the EHS pipeline](https://trello.com/c/87CyAnik/826-add-uwdormroomnumber-back-into-the-ehs-pipeline)

bin/uw-ehs-report/transform
jccraft-uw: OK, I'm going to stay with how I have it. "1" and "0".
Pandas is smart. It interprets the newly created column as a float. However, we want the value written to the CSV as a string. At the very beginning when we load the CSV we explicitly say that we want every column brought in as a string ... starting on line 17 note the .astype("string")  (twice because this was Tom's workaround to the Pandas change that broke barcodes a few weeks ago.)
Therefore, I want the symptomatic_when_tested to be a string too. I could let Pandas create it as a float and then change the type, or I could just simply assign string values as I have already done.

joverlee521: Makes sense to leave as is then! Could you add a similar comment that REDCap expects 1/0 values?

jccraft-uw: Added in the new FIXUP commit.

kschwabefry (approved):

bin/uw-ehs-report/transform
kairstenfay: In that case, for internal consistency of using string values of '0' and '1', it may be clearer to cast 't' and 'f' to '1' and '0' in line 172 (rather than 1 and 0).

bin/uw-ehs-report/transform
```diff
@@ -91,10 +91,17 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+        'record_id',
+        'barcode',
+        'illness_kiosk',
+        'illness_swabsend'
         ]]
 
+    # Set symptomatic_when_tested on encounters.
+    # Use '1' and '0' values to be consistent with the other boolean values we send to REDCap.
+    encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
+    encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
```
tsibley, line 103: It looks to me like this logic won't do the right thing.

Imagine a record where:

```
illness_kiosk = yes
illness_swabsend = no
```

won't both lines match that record and thus set `symptomatic_when_tested` to 1 and then immediately set it again to 0?

I think the correct logic (in pseudo-code) would be:

```
illness_kiosk == yes  or illness_swabnsend == yes → symptomatic_when_tested = 1
illness_kiosk == no  and illness_swabnsend == no  → symptomatic_when_tested = 0
```

…except that in Pandas both conditions will have to handle nulls/NA values properly, which I'm not sure will happen with the `==` operator.

Or maybe I'm missing something?

bin/uw-ehs-report/transform
jccraft-uw: Then I would have to replace the 't' and 'f' values before joining the data.

bin/uw-ehs-report/transform
jccraft-uw: There should not be both sets of tests in the same encounter. The PT gets either a kiosk test or a swab and send test. If the instrument isn't filled out the value for the illness_* question is empty, not 'no'.

PR #93, opened 2020-10-23
fatigue: Modify to suppress multiple, repeated warnings

kairstenfay:
Previously, fatigue only specially handled suppressing repeated warnings
of jobs that exit with an error. Modify fatigue to also suppress jobs
with any sterr output. Sterr output is generated at the "warning" and
"error" log levels, so this will reduce the warnings alert spam we
receive via email and in the #id3c-alerts channel.

### DIFF
--- a/bin/fatigue
+++ b/bin/fatigue
@@ -105,7 +105,7 @@ main() {
     # Run the given command, with output sent to the new state file.
     "$@" &>"$new_state" && exited=0 || exited=$?
 
-    if [[ $exited -ne 0 ]]; then
+    if [[ $exited -ne 0 || -s "$new_state" ]]; then
         if [[ ! -e "$old_state" || "$(hash-state "$new_state")" != "$(hash-state "$old_state")" ]]; then
             # A new error! Save it for later and exit with error.
             cat "$new_state"
--- a/bin/fatigue
+++ b/bin/fatigue
@@ -105,7 +105,7 @@ main() {
     # Run the given command, with output sent to the new state file.
     "$@" &>"$new_state" && exited=0 || exited=$?
 
-    if [[ $exited -ne 0 || -s "$new_state" ]]; then
+    if [[ $exited -ne 0 || (-s "$new_state" && $quiet -eq 1) ]]; then
         if [[ ! -e "$old_state" || "$(hash-state "$new_state")" != "$(hash-state "$old_state")" ]]; then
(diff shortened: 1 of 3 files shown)


kairstenfay:
I just tested this and realized that any stdout ouptut will result in a new fatigue state. I'm not sure if that's what we want.

kairstenfay:
[Make #id3c-alerts warnings less spammy](https://trello.com/c/fOFwif8b/669-make-id3c-alerts-warnings-less-spammy)

tsibley:
As part of thinking about this, I took a step back and laid out the existing behaviour of `fatigue` vs. `fatigue --quiet`, both ±chronic, in a [decision table](https://raw.githubusercontent.com/tsibley/fatigue/main/using-with-chronic.rst). I believe the behaviour we want, as described by this PR, is obtainable currently via `fatigue --quiet chronic -e …`. That said, we may still want to modify `fatigue` to support it directly.

I think this also makes clear that there's a compelling argument to be made for abstracting over `chronic` and/or `fatigue` by wrapping them in some command¹ with a clearer set of behavioural flags that maybe compose more intuitively e.g.

* `--suppress-output={on-success | on-repeat | on-repeat-failure}`
* `--stderr-is-failure`

¹ A new command or maybe just a new version of `fatigue`.

tsibley:
Some observations:

- As you note, this will track old state for successful commands that have any output. An invisible behaviour change unless you're looking at state files to see what's still failing.
- The only visible behaviour change happens when `--quiet` is used to suppress repeat output.

This suggests that the added conditional, and thus old state tracking on success, could be scoped to times when `--quiet` is also used.

kairstenfay:
> Some observations:
> 
>     * As you note, this will track old state for successful commands that have any output. An invisible behaviour change unless you're looking at state files to see what's still failing.
> 
>     * The only visible behaviour change happens when `--quiet` is used to suppress repeat output.
> 
> 
> This suggests that the added conditional, and thus old state tracking on success, could be scoped to times when `--quiet` is also used.

I just pushed a commit that I believe addresses these notes of yours @tsibley. Please let me know how the changes look if you get a chance to review it. 

However, I also am thinking of implementing your previous suggestion of an `--suppress-output={on-success | on-repeat | on-repeat-failure}` option (new version of `fatigue`) that may just rewrite these commits completely. However, I won't have time to finish that in the next couple of weeks due to my switching to Nextstrain.

PR #94, opened 2020-10-26
Add utility script to mint barcodes in batch

jccraft-uw:


bin/mint-barcodes-in-batch
```diff
@@ -0,0 +1,66 @@
...
+    required = True)
+
+@click.option("--file_prefix", type=str,
+    help = "The prefix (including path) of the label PDF files to create",
+    required = True)
+
+@click.option("--sheets_per_file", type=int,
+    help = "The maximum number of sheets to include in a single PDF file",
+    default = 20,
+    required = False)
```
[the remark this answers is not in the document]
kairstenfay: It might also be nice to shorten some of these option names. For example, "number of sheets" to "sheets". Other suggestions are less obvious to me.

bin/mint-barcodes-in-batch
```diff
@@ -0,0 +1,66 @@
...
+
+    sheets_remaining = number_of_sheets
+    iterations = 1
+
+    try:
+        while sheets_remaining > 0:
+            if sheets_remaining > sheets_per_file:
+                num_sheets = sheets_per_file
+            else:
+                num_sheets = sheets_remaining
```
kairstenfay, line 51: This could be simplified with `num_sheets = min(sheets_remaining, sheets_per_file)`

bin/mint-barcodes-in-batch
```diff
@@ -0,0 +1,66 @@
...
+
+            subprocess.run(["pipenv", "run", "id3c", "identifier", "mint",
+                    f"{identifier_set}", f"{number_per_sheet * num_sheets}", "--quiet", "--labels",
+                    f"{file_prefix}{identifier_set}_{num_sheets}_sheets_{date.today()}_{iterations}.pdf"],
+                    capture_output=True, check=True)
+            sheets_remaining = sheets_remaining - num_sheets
+            iterations = iterations + 1
```
kairstenfay, line 58: In python, we can write:
```py
sheets_remaining -= num_sheets
iterations += 1
```

kairstenfay (commented):
Thanks for writing this! Barcode generation is never fun, and I haven't even attempted to tackle the splitting of PDFs into fewer sheets.

As I mentioned in Slack, given that this is the script you're already using to mint barcodes, it's OK to merge as-is so that we can capture the current state of production.

However, I left some style comments that would make this a more pythonic CLI program.

bin/mint-barcodes-in-batch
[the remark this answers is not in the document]
jccraft-uw: Thanks for pointing out the hypen vs underscore part. I've changed that and made the option names shorter.

bin/mint-barcodes-in-batch
jccraft-uw: That is the coolest thing ever. min ... what a really useful builtin function. Thanks for that.

bin/mint-barcodes-in-batch
jccraft-uw: Thanks. I gave up when "iterations++" wasn't recognized.
I'm using what you wrote.

PR #96, opened 2020-10-28
ROR: Fix config path

joverlee521:


kairstenfay (approved):
Locally, the base dir now points to the top level of my backoffice repo.

tsibley (approved):

PR #98, opened 2020-10-29
Update to season 3 manifest

kairstenfay:
Update the `update-unattended` specimen manifests cron job environment
variables to pull from the new, season 3 manifest sheet.

Note that the lab does not anticipate updating the season 2 manifest save any barcode corrections. If these arise, the plan is to handle them manually.

Depends on https://github.com/seattleflu/specimen-manifests/pull/22

### DIFF
--- a/id3c-production/env.d/aliquot-manifest-processing/CONFIG
+++ b/id3c-production/env.d/aliquot-manifest-processing/CONFIG
@@ -1,1 +1,1 @@
-/opt/specimen-manifests/season-2-AWS-S3.yaml
+/opt/specimen-manifests/season-3-AWS-S3.yaml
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Account for slimmer Y3 aliquot manifest sheet](https://trello.com/c/0ij7bOnA/869-account-for-slimmer-y3-aliquot-manifest-sheet)

kschwabefry (approved):
Should this be deployed on Sunday, once the lab switches to the new filename?

kairstenfay:
> Should this be deployed on Sunday, once the lab switches to the new filename?

Yes, or potentially early Monday. I unfortunately don't see a better solution given the status quo. It would be nice in the future to change our system so we're able to deploy these PRs before the aliquoting sheet exists on S3.

tsibley (approved):

tsibley:
> It would be nice in the future to change our system so we're able to deploy these PRs before the aliquoting sheet exists on S3.

Agreed. One wrinkle here is that we would still _sometimes_ want to know if the manifest isn't there, as it has occasionally disappeared on us due to uploading mistakes.

PR #102, opened 2020-12-01
Update UW retrospecive manifest config with season 3 manifests

joverlee521:
Forgot to update this config when we transitioned to the new season
3 manifests.


jccraft-uw:
It's too bad there's a separate manifest for the same thing; it's too bad this process can't use the same manifest that the manifest processing uses.

jccraft-uw (approved):

kairstenfay (approved):

PR #103, opened 2020-12-02
Add cronjob for uw reopening offers

jccraft-uw:
As I noted as comments in the cron file, run this job at :15 and :45 after the hour.

redcap-det uw-reopening runs at  /10
etl fhir runs at /5
refresh-materialized-view shipping fhir_questionnaire_responses_v1 runs at /5

The offer job uses data from all 3 of those other jobs. Run this at :15 to use the :00 redcap-det uw-reopening run and at :45 to use the :30 redcap-det uw-reopening run.

Space the two runs out to give the other jobs a few times to run between.

### DIFF
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -121,3 +121,8 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 # ingested FHIR documents in the view
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.fhir_questionnaire_responses_v1" fatigue --quiet pipenv run id3c refresh-materialized-view shipping fhir_questionnaire_responses_v1 --commit
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.scan_encounters_v1"              fatigue --quiet pipenv run id3c refresh-materialized-view shipping scan_encounters_v1 --commit
+
+# Make UW Reopening testing offers at :15 and :45
+# This job uses data created by redcap-det uw-reopening, etl fhir, and refresh-materialized-view shipping fhir_questionnaire_responses_v1
+# Run at :15 to use the :00 uw-reopening run and :45 to use the :30 uw-reopening run
+15,45  * * * * ubuntu promjob "id3c offer-uw-testing" fatigue --quiet envdir $ENVD/redcap-sfs pipenv run id3c offer-uw-testing --commit


jccraft-uw:
[Make the cron job for the UW quota job](https://trello.com/c/pK9s4D7d/917-make-the-cron-job-for-the-uw-quota-job)

crontabs/id3c-production
```diff
@@ -121,3 +121,8 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 # ingested FHIR documents in the view
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.fhir_questionnaire_responses_v1" fatigue --quiet pipenv run id3c refresh-materialized-view shipping fhir_questionnaire_responses_v1 --commit
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.scan_encounters_v1"              fatigue --quiet pipenv run id3c refresh-materialized-view shipping scan_encounters_v1 --commit
+
+# Make UW Reopening testing offers at :15 and :55
+# This job uses data created by redcap-det uw-reopening, etl fhir, and refresh-materialized-view shipping fhir_questionnaire_responses_v1
+# Run at :15 to use the :00 uw-reopening run and :55 to use the :40 uw-reopening run
+15,55  * * * * ubuntu promjob "id3c offer-uw-testing" fatigue --quiet $ENVD/redcap-sfs pipenv run id3c offer-uw-testing --commit
```
kschwabefry, line 128: Could we run this every half hour to space the runs out evenly? Perhaps  :15 and :45. The uw-reopening run is scheduled for every ten minutes on so it doesn't seem like we would have to target specific runs of that.

jccraft-uw: We need the 3 jobs to run to get all the data updated before we run the quota job. If we run quota before the materialization refresh has finished, for example, we risk allocating duplicate spots to the same queue entries.

jccraft-uw: Let's see here .... We want to run twice an hour. It if makes things saner, I'll space it out 30 minutes between runs. Running at :15 and :45 sounds fine with me. I'll make that change.

kschwabefry: Ok! Thanks Chris. I don't think I was following how :15 :55 minimized risk of assigning slots before the other jobs were complete was all.

kschwabefry (approved):

crontabs/id3c-production
```diff
@@ -121,3 +121,8 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 # ingested FHIR documents in the view
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.fhir_questionnaire_responses_v1" fatigue --quiet pipenv run id3c refresh-materialized-view shipping fhir_questionnaire_responses_v1 --commit
 */5 * * * * ubuntu promjob "refresh-materialized-view: shipping.scan_encounters_v1"              fatigue --quiet pipenv run id3c refresh-materialized-view shipping scan_encounters_v1 --commit
+
+# Make UW Reopening testing offers at :15 and :45
+# This job uses data created by redcap-det uw-reopening, etl fhir, and refresh-materialized-view shipping fhir_questionnaire_responses_v1
+# Run at :15 to use the :00 uw-reopening run and :45 to use the :30 uw-reopening run
+15,45  * * * * ubuntu promjob "id3c offer-uw-testing" fatigue --quiet $ENVD/redcap-sfs pipenv run id3c offer-uw-testing --commit
```
tsibley, line 128: Looks like this is missing the `envdir` command before its `$ENVD/…` argument.

tsibley (approved):

tsibley (changes requested):
Whoops, fat-fingered the review status.

crontabs/id3c-production
jccraft-uw: That would certainly make things work better. Added.

jccraft-uw:
> Whoops, fat-fingered the review status.

Thanks for calling it out. I fixed that and am deploying now and will babysit it.

jccraft-uw:
Deployed.

PR #104, opened 2020-12-03
Silence alerts for Hap1 and exp samples

kairstenfay:
Update our logging configuration to silence yet another test sample that
isn't meant to end up in id3c.

### DIFF
--- a/id3c-production/logging.yaml
+++ b/id3c-production/logging.yaml
@@ -39,7 +39,7 @@ filters:
     name: id3c.cli.command.etl.presence_absence
     levelname: WARNING
     msg:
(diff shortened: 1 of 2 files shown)


jccraft-uw (approved):
Thanks a lot for doing this.

kschwabefry:
Thanks Kairsten! I see another alert last night  (12/2 at 7:45 PM) for "exp" without prefix and underscore, which could also be filtered out of our warnings. Do you mind adding that as well while you're updating this filter?

kairstenfay:
> Thanks Kairsten! I see another alert last night (12/2 at 7:45 PM) for "exp" without prefix and underscore, which could also be filtered out of our warnings. Do you mind adding that as well while you're updating this filter?

Updated! Please take a look when you get a chance.

kschwabefry (approved):

PR #105, opened 2020-12-03
Musher metrics

tsibley:
To be deployed with or after https://github.com/seattleflu/husky-musher/pull/7.


husky-musher/uwsgi.ini
```diff
@@ -27,3 +27,9 @@ enable-threads = true
 # 1.5× the default buffer size for request headers to accommodate large user
 # information passed in the headers from Shibboleth/Apache.
 buffer-size = 6144
+
+# Setup a clean Prometheus metrics collection dir every time; required since
+# we're collecting metrics across multiple processes.  Set env here instead of
+# in the envdir so that we can use it in the exec-pre-app line.
+env = prometheus_multiproc_dir=/run/uwsgi/%(deb-confnamespace)/%(deb-confname)/prometheus
+exec-pre-app = mkdir -p "$prometheus_multiproc_dir" && rm -vf "$prometheus_multiproc_dir"/*
```
kschwabefry, line 35: Let's see. So old metrics get wiped when we use the ini file... does that happen when uwsgi is restarted (hopefully not reloaded)? I'm on board with not retaining metrics indefinitely but I wonder if there's a way we can delete them on the basis of age instead of on a uwsgi restart. Comparing metrics before and after a restart may be useful.

tsibley: `exec-pre-app` happens on restart and reload. This metrics collection dir is only for aggregation/coordination across multiple processes and is by design intended to be very transient/temporary. The Prometheus server will scrape the metrics frequently (currently at the pretty conservative rate of every minute; could increase to every 30s or 15s) and store them longer term for the configured retention time ([default is 15d](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects), which we use).

kairstenfay (approved):

kschwabefry (approved):

PR #106, opened 2020-12-07
Switchboard enhancement

jccraft-uw:
These are a couple of small changes to make updating the Switchboard a bit smoother.

### DIFF
--- a/crontabs/scan-switchboard
+++ b/crontabs/scan-switchboard
@@ -12,4 +12,4 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 
 
 # Refresh the SCAN Switchboard SQLite database every 10 minutes
-*/10 * * * * ubuntu promjob "switchboard refresh" chronic fatigue envdir $ENVD/redcap-sfs/ envdir $ENVD/redcap-scan/ pipenv run make -BC /opt/scan-switchboard
+*/10 * * * * ubuntu promjob "switchboard refresh" chronic fatigue envdir $ENVD/redcap-sfs/ envdir $ENVD/redcap-scan/ flock /opt/scan-switchboard/data/record-barcodes.ndjson pipenv run make -BC /opt/scan-switchboard
--- a/crontabs/scan-switchboard
+++ b/crontabs/scan-switchboard
@@ -11,5 +11,5 @@ ENVD=/opt/backoffice/id3c-production/env.d
 XDG_RUNTIME_DIR=/home/ubuntu/run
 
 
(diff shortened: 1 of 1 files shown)


jccraft-uw:
[scan-switchboard: make loading data more robust](https://trello.com/c/zwUaQ9fq/882-scan-switchboard-make-loading-data-more-robust)

crontabs/scan-switchboard
```diff
@@ -11,5 +11,5 @@ ENVD=/opt/backoffice/id3c-production/env.d
 XDG_RUNTIME_DIR=/home/ubuntu/run
 
 
-# Refresh the SCAN Switchboard SQLite database every 10 minutes
-*/10 * * * * ubuntu promjob "switchboard refresh" chronic fatigue envdir $ENVD/redcap-sfs/ envdir $ENVD/redcap-scan/ pipenv run make -BC /opt/scan-switchboard
+# Refresh the SCAN Switchboard SQLite database every 30 minutes
+*/30 * * * * ubuntu promjob "switchboard refresh" chronic fatigue envdir $ENVD/redcap-sfs/ envdir $ENVD/redcap-scan/ flock /opt/scan-switchboard/data/record-barcodes.ndjson pipenv run make -BC /opt/scan-switchboard
```
tsibley, line 15: All our other invocations of `flock` use its `--no-fork` / `-F` option to continue the exec-chain instead of forking the chain and using child processes. I think it'd be best to continue to do that here too.

tsibley (changes requested):

crontabs/scan-switchboard
kschwabefry: There's text in the switchboard UI https://backoffice.seattleflu.org/switchboard/sfs-redcap that says "updated every 5 minutes" - which I think comes from https://github.com/seattleflu/scan-switchboard/blob/21cb1d6f0f3618aa72a4b8502faffd238c543002/datasette.yaml#L8 . @jccraft-uw do you mind making that change also to reflect the 30 minute refresh schedule?

jccraft-uw: > All our other invocations of `flock` use its `--no-fork` / `-F` option to continue the exec-chain instead of forking the chain and using child processes. I think it'd be best to continue to do that here too.

Thanks for seeing that I missed the "-F", @tsibley . I've added it in.
@kschwabefry I've made the "updated every ... " change in the branch on the other repro. Thanks for seeing that.

jccraft-uw:
Deployed.

PR #107, opened 2020-12-08
Import HCT data to EHS REDCap in batches

kschwabefry:
Import Husky Coronavirus Testing data into the EHS REDCap transfer
project in batches of 10,000 records. Uploading the full dataset
(currently 24k records) was causing the import into REDCap to fail.

### DIFF
--- a/bin/uw-ehs-report/import-to-uw-ehs-transfer-redcap
+++ b/bin/uw-ehs-report/import-to-uw-ehs-transfer-redcap
@@ -2,6 +2,8 @@
 import argparse
 import os
 import json
+import math
+import numpy as np
 import pandas as pd
 import requests
 import logging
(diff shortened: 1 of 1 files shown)


jccraft-uw (approved):
This looks great, @kschwabefry .
I generated the CSV from Prod. It has 24,980 rows, with the first being the header. So, that's 24,979 rows of data.

array_split returns these batches:
4 with 1,000 rows
21 with 999 rows
That adds up  ... 24,979

The only negative is that it takes longer to import the records into REDCap. That's the reality of doing them in batches instead of doing them all at once, but that's what we need to do.

kschwabefry:
Updated to change batch size from 1k records to 10k records. This shaves off a few minutes, and should still be small enough for redcap to handle - as we didn't start running into issues until we hit 24k records.

kschwabefry:
Deployed.

PR #108, opened 2020-12-12
id3c-production: drop duplicate HCT barcodes for EHS transfer

kschwabefry:
Before import data into the EH&S HCT transfer project, drop duplicate
barcodes from the dataset.


joverlee521 (approved):
Thanks for the quick fix @kschwabefry!

kschwabefry:
Deployed.

PR #111, opened 2020-12-28
return-of-results: Exit if id3c export fails

kairstenfay:
Exit the return-of-results (RoR) job if the CSV export from id3c fails
(e.g. because the database is unavailable). Previously, we were not
killing the RoR job on a failed `export-id3c-return-results` command due
to our use of process substitution. I.e., as @tsibley pointed out, "when
`command` fails in `<(command)`, it doesn't propagate."

Explicitly kill the current PID when our id3c export command exits with
a non-zero exit code. This halts the entire RoR job, preventing us from
unintentionally uploading an empty CSV and overrwriting the
(unversioned) data on securelink's S3 bucket with an empty file. This
happened last month, causing the RoR job to regenerate PDFs for every
single result, which at our current program's pace, would have taken
multiple days to complete.

### DIFF
--- a/bin/return-of-results/generate-results-csv
+++ b/bin/return-of-results/generate-results-csv
@@ -23,7 +23,7 @@ main() {
     # Change to the ./bin/return-of-results directory in the backoffice checkout
     cd "$(dirname "$0")"
 
-    ./transform <(./export-redcap-scan) <(./export-redcap-sfs-longitudinal) <(./export-id3c-return-results)
+    ./transform <(./export-redcap-scan) <(./export-redcap-sfs-longitudinal) <(./export-id3c-return-results || kill $$)
 }
 
 print-help() {


kairstenfay:
[Exit return of results job with non-zero error code if id3c-export returns no results](https://trello.com/c/WdStlXdh/909-exit-return-of-results-job-with-non-zero-error-code-if-id3c-export-returns-no-results)

jccraft-uw:
> Exit the return-of-results (RoR) job if the CSV export from id3c fails
> (e.g. because the database is unavailable). Previously, we were not
> killing the RoR job on a failed `export-id3c-return-results` command due
> to our use of process substitution. I.e., as @tsibley pointed out, "when
> `command` fails in `<(command)`, it doesn't propagate."
> 
> Explicitly kill the current PID when our id3c export command exits with
> a non-zero exit code. This halts the entire RoR job, preventing us from
> unintentionally uploading an empty CSV and overrwriting the
> (unversioned) data on securelink's S3 bucket with an empty file. This
> happened last month, causing the RoR job to regenerate PDFs for every
> single result, which at our current program's pace, would have taken
> multiple days to complete.

Is `<(./export-id3c-return-results || kill $$)`
killing the process because the command exited with a non-zero exit code or because the command returned nothing?
Is the exit code of that process being checked?

kairstenfay:
> Is `<(./export-id3c-return-results || kill $$)`
> killing the process because the command exited with a non-zero exit code or because the command returned nothing?
> Is the exit code of that process being checked?

It kills the process because of the non-zero exit code of the command. I.e. the exit code of the process is being checked. I am having trouble finding documentation that supports this, but it's possible to see that the `generate` script does not `kill` the current PID on a zero exit code that returns nothing by modifying the `export-id3c-return-results` script to simply "exit 0"

jccraft-uw:
> > Is `<(./export-id3c-return-results || kill $$)`
> > killing the process because the command exited with a non-zero exit code or because the command returned nothing?
> > Is the exit code of that process being checked?
> 
> It kills the process because of the non-zero exit code of the command. I.e. the exit code of the process is being checked. I am having trouble finding documentation that supports this, but it's possible to see that the `generate` script does not `kill` the current PID on a zero exit code that returns nothing by modifying the `export-id3c-return-results` script to simply "exit 0"

Gotcha. I didn't know anything about process substitution, so I did some Googling and read https://unix.stackexchange.com/questions/128560/how-do-i-capture-the-exit-code-handle-errors-correctly-when-using-process-subs
On there I saw "Processes in process substitution are asynchronous: the shell launches them and then doesn't give any way to detect when they die. So you won't be able to obtain the exit status." That led to me to ask.

kschwabefry (approved):
Thanks Kairsten! Per your comment above, here's a documentation link for OR operator: https://www.gnu.org/software/bash/manual/bash.html#Lists

"command2 is executed if, and only if, command1 returns a non-zero exit status."

kairstenfay:
> Thanks Kairsten! Per your comment above, here's a documentation link for OR operator: https://www.gnu.org/software/bash/manual/bash.html#Lists
> 
> "command2 is executed if, and only if, command1 returns a non-zero exit status."

Thanks so much, Kristen!

tsibley:
Why does only one of the three process substitutions in that line get this treatment?

kairstenfay:
> Why does only one of the three process substitutions in that line get this treatment?

Good point. I would say because only one of the three process substitutions in that line gave us a headache when it exited with a non-zero exit code. Are you suggesting adding an `|| kill $$` statement to the other two?

tsibley:
Nod. My default expectation, in the absence of mitigating context, is that fixes address whole classes of bugs, rather than single instances. Maybe, after consideration, the other instances don't matter as much, but I do think worth at least considering if the other process substitutions can cause similar headaches if they fail. If so, seems like the same fix is warranted.

kairstenfay:
> Nod. My default expectation, in the absence of mitigating context, is that fixes address whole classes of bugs, rather than single instances. Maybe, after consideration, the other instances don't matter as much, but I do think worth at least considering if the other process substitutions can cause similar headaches if they fail. If so, seems like the same fix is warranted.

I just tried to run `generate-results-csv` using `|| kill $$` in the other processes substituted in the script, but it appears they terminate on their own when they encounter an error (e.g. a ValueError, RuntimeError, or ModuleNotFoundError). I get output like:
```
./generate-results-csv: line 26: kill: ({PID}) - No such process
```

Is this the correct interpretation? If so, I can add a comment to the `generate-results-csv` script explaining why only one of the process substitutions gets the `|| kill $$` treatment.

tsibley:
Hmm. I'm not quite sure I follow. Can you paste a diff of what you tested so we're sure we're on the same page there? Let me also explain my understanding so far, and maybe we can see where the misunderstanding is?

All the processes we're talking about terminate on their own when they encounter an uncaught error. However, the issue addressed by this PR is that process substitution in Bash doesn't bubble errors up to the parent process _even under `set -e`_. The workaround of using `x || kill $$` in the process substitution means that when `x` fails, the child process signals the parent process (`$$`) to terminate with error. This happens outside of the normal parent/child status handling.

Without seeing what exactly you were testing, I'm not sure about the "No such process" you're seeing. My best guess would be a race condition where the child process tries to signal the parent after the parent has already exited. (But I would expect the parent to wait for the child to exit before exiting itself.)

kairstenfay:
> Hmm. I'm not quite sure I follow. Can you paste a diff of what you tested so we're sure we're on the same page there? Let me also explain my understanding so far, and maybe we can see where the misunderstanding is?
> 
> All the processes we're talking about terminate on their own when they encounter an uncaught error. However, the issue addressed by this PR is that process substitution in Bash doesn't bubble errors up to the parent process _even under `set -e`_. The workaround of using `x || kill $$` in the process substitution means that when `x` fails, the child process signals the parent process (`$$`) to terminate with error. This happens outside of the normal parent/child status handling.
> 
> Without seeing what exactly you were testing, I'm not sure about the "No such process" you're seeing. My best guess would be a race condition where the child process tries to signal the parent after the parent has already exited. (But I would expect the parent to wait for the child to exit before exiting itself.)

@tsibley Thanks for writing this out. We are on the same page about the `x || kill $$` workaround within process substitution. 

I'm not sure about the race condition, part, though. Here's what I've tried.

1. Running `./bin/return-of-results/generate-results-csv` with no id3c environment and no database connection. I get this output:
```
Traceback (most recent call last):
  File "./export-redcap-scan", line 7, in <module>
    from id3c.cli import redcap
ModuleNotFoundError: No module named 'id3c'
Traceback (most recent call last):
  File "./export-redcap-sfs-longitudinal", line 32, in <module>
    from id3c.cli.redcap import Project
ModuleNotFoundError: No module named 'id3c'
Traceback (most recent call last):
  File "./export-redcap-sfs-classic", line 31, in <module>
    from id3c.cli.redcap import Project
ModuleNotFoundError: No module named 'id3c'
ERROR:  relation "shipping.return_results_v3" does not exist
LINE 1: COPY  ( select * from shipping.return_results_v3 ) TO STDOUT...
                              ^
Terminated
(base) kairsten:backoffice$ Traceback (most recent call last):
  File "./transform", line 265, in <module>
    scan_redcap_data = parse_scan_redcap(args.scan_redcap_data)
  File "./transform", line 19, in parse_scan_redcap
    .astype("string")
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/generic.py", line 5702, in astype
    result = pd.concat(results, axis=1, copy=False)
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/reshape/concat.py", line 281, in concat
    sort=sort,
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/reshape/concat.py", line 329, in __init__
    raise ValueError("No objects to concatenate")
ValueError: No objects to concatenate
```

2. Running the same script in step one but with `x || kill $$` added to each process substitution, e.g.
```diff
-    ./transform <(./export-redcap-scan) <(./export-redcap-sfs-longitudinal) <(./export-redcap-sfs-classic) <(./export-id3c-return-results || kill $$)
+    ./transform <(./export-redcap-scan || kill $$) <(./export-redcap-sfs-longitudinal || kill $$) <(./export-redcap-sfs-classic || kill $$) <(./export-id3c-return-results || kill $$)
```

The output changes slightly. Namely, there's a `Terminated` printed after `./export-redcap-scan` fails. The "No such process" statements are also printed.
```
Traceback (most recent call last):
  File "./export-redcap-scan", line 7, in <module>
    from id3c.cli import redcap
ModuleNotFoundError: No module named 'id3c'
Terminated
(base) kairsten:backoffice$ Traceback (most recent call last):
  File "./export-redcap-sfs-classic", line 31, in <module>
    from id3c.cli.redcap import Project
ModuleNotFoundError: No module named 'id3c'
ERROR:  relation "shipping.return_results_v3" does not exist
LINE 1: COPY  ( select * from shipping.return_results_v3 ) TO STDOUT...
                              ^
./bin/return-of-results/generate-results-csv: line 26: kill: (12692) - No such process
./bin/return-of-results/generate-results-csv: line 26: kill: (12692) - No such process
Traceback (most recent call last):
  File "./export-redcap-sfs-longitudinal", line 32, in <module>
    from id3c.cli.redcap import Project
ModuleNotFoundError: No module named 'id3c'
./bin/return-of-results/generate-results-csv: line 26: kill: (12692) - No such process
Traceback (most recent call last):
  File "./transform", line 265, in <module>
    scan_redcap_data = parse_scan_redcap(args.scan_redcap_data)
  File "./transform", line 19, in parse_scan_redcap
    .astype("string")
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/generic.py", line 5702, in astype
    result = pd.concat(results, axis=1, copy=False)
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/reshape/concat.py", line 281, in concat
    sort=sort,
  File "/home/kairsten/miniconda3/lib/python3.7/site-packages/pandas/core/reshape/concat.py", line 329, in __init__
    raise ValueError("No objects to concatenate")
ValueError: No objects to concatenate
```

tsibley:
Ah, I see. Here's what happening. It is indeed the race condition I guessed at, although now I understand why my expectation that the parent would wait for the child was faulty.

The four `|| kill $$` in the process substitutions handle bubbling up of errors in each by killing the parent.

The first to fail will successfully terminate the parent process, and that's why you see `Terminated` in the fifth line of output.

Any others that fail but don't win the race to be first will fail to `kill` the parent, since it's not around anymore, which is why you see three lines of:

```
./bin/return-of-results/generate-results-csv: line 26: kill: (12692) - No such process
```

In your example, all four child processes fail because they can't load `id3c`.

However, in normal circumstances, terminating the parent process due to failure of just one child may in many cases also _cause_ the other children to error (and thus attempt to `kill`), because they were writing to the parent process, which was reading from them before it went away suddenly.

PR #112, opened 2020-12-29
Add cron job to refresh uw materialized view

jccraft-uw:
Do not deploy until corresponding ID3C customizations are deployed.
That is https://github.com/seattleflu/id3c-customizations/pull/174


jccraft-uw:
Deployed.

PR #113, opened 2021-01-05
refresh-database: Exclude comments on extensions

kairstenfay:
Remove comments on extensions from the TOC when refreshing a target
database with a copy of the production database.

@tsibley pointed out that comments on extensions require a true
superuser, but RDS doesn't offer us that. Therefore, filter out comments
on extensions while preserving other comments (instead of using the
nuclear `--no-comment` option available in Postgres 11).

### DIFF
--- a/dev/refresh-database
+++ b/dev/refresh-database
@@ -239,12 +239,12 @@ suppress-ignorable-errors() {
 
         s/(?-x:^pg_restore: .+? COMMENT EXTENSION (.+?)\s*)\n
           (?-x:^pg_restore: .+? ERROR:  must be owner of extension .+?)\n
-          (?-x:^Command was: COMMENT ON EXTENSION .+?)\n+
+          (?-x:^\s*Command was: COMMENT ON EXTENSION .+?)\n+
          /COMMENT EXTENSION \1 (failure due to RDS permissions ignored)\n/xmg;
 
         s/(?-x:^pg_restore: .+? TABLE DATA spatial_ref_sys rdsadmin)\n
           (?-x:^pg_restore: .+? ERROR:  permission denied for relation spatial_ref_sys)\n
-          (?-x:^Command was: COPY public\.spatial_ref_sys .+?)\n+
+          (?-x:^\s*Command was: COPY public\.spatial_ref_sys .+?)\n+
          /TABLE DATA spatial_ref_sys (failure due to RDS permissions ignored)\n/xmg;
     '
 }
--- a/dev/refresh-database
+++ b/dev/refresh-database
@@ -106,8 +106,10 @@ main() {
     log "Restoring filtered roles"
     psql --no-psqlrc < production-roles-filtered.sql |& suppress-ignorable-errors
 
+    log "Removing comments on extensions from the TOC"
+    pg_restore --list production.pgdb | grep -v 'COMMENT - EXTENSION' > production.toc
     log "Restoring database dump"
-    pg_restore --dbname "$PGDATABASE" production.pgdb |& suppress-ignorable-errors || true
+    pg_restore --dbname "$PGDATABASE" --use-list production.toc production.pgdb |& suppress-ignorable-errors || true
 
     log "Finished in $(elapsed-time)"
 }
@@ -236,11 +238,6 @@ suppress-ignorable-errors() {
     # process so that real, unexpected errors aren't lost in familiar noise.
     perl -000pe '
         s/^ERROR:  role (.+?) already exists/CREATE ROLE \1 (already exists, skipped)/mg;
-
-        s/(?-x:^pg_restore: .+? COMMENT EXTENSION (.+?)\s*)\n
-          (?-x:^pg_restore: .+? ERROR:  must be owner of extension .+?)\n
-          (?-x:^\s*Command was: COMMENT ON EXTENSION .+?)\n+
-         /COMMENT EXTENSION \1 (failure due to RDS permissions ignored)\n/xmg;
 
         s/(?-x:^pg_restore: .+? TABLE DATA spatial_ref_sys rdsadmin)\n
           (?-x:^pg_restore: .+? ERROR:  permission denied for relation spatial_ref_sys)\n
--- a/dev/refresh-database
+++ b/dev/refresh-database
@@ -107,7 +107,8 @@ main() {
     psql --no-psqlrc < production-roles-filtered.sql |& suppress-ignorable-errors
 
     log "Removing comments on extensions from the TOC"
-    pg_restore --list production.pgdb | grep -v 'COMMENT - EXTENSION' > production.toc
+    pg_restore --list production.pgdb \
+        | grep -vE 'COMMENT - EXTENSION|TABLE DATA public spatial_ref_sys rdsadmin' > production.toc
     log "Restoring database dump"
     pg_restore --dbname "$PGDATABASE" --use-list production.toc production.pgdb |& suppress-ignorable-errors || true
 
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Refresh id3c testing instance with new data](https://trello.com/c/bZhtlDMq/953-refresh-id3c-testing-instance-with-new-data)

dev/refresh-database
```diff
@@ -106,8 +106,10 @@ main() {
     log "Restoring filtered roles"
     psql --no-psqlrc < production-roles-filtered.sql |& suppress-ignorable-errors
 
+    log "Removing comments on extensions from the TOC"
+    pg_restore --list | grep -v 'COMMENT - EXTENSION' > production.toc
```
tsibley, line 110: `pg_restore` still needs to be given the dump file as an argument.

tsibley (commented):
When reviewing this PR, I noticed this existing code to make the error output nicer:

https://github.com/seattleflu/backoffice/blob/70b67f7cee6ede75f28b109d0a459b1a0fa22d45/dev/refresh-database#L240-L243

From the [log you pasted in Slack](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1609808681024700?thread_ts=1609808492.024200&cid=C0173B4S1LY), it looks like that stopped working, maybe due to changes in error formatting?

dev/refresh-database
kairstenfay: Ah, thanks for catching that.

kairstenfay:
> When reviewing this PR, I noticed this existing code to make the error output nicer:
> 
> https://github.com/seattleflu/backoffice/blob/70b67f7cee6ede75f28b109d0a459b1a0fa22d45/dev/refresh-database#L240-L243
> 
> From the [log you pasted in Slack](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1609808681024700?thread_ts=1609808492.024200&cid=C0173B4S1LY), it looks like that stopped working, maybe due to changes in error formatting?

I noticed that, too, and was surprised to see that the silencing of noisy logs no longer worked. I forgot to look into it, but I'll see if I can re-align that code with the new error formatting now.

dev/refresh-database
kairstenfay: Updated in the newest commit.

kairstenfay (commented):
> > When reviewing this PR, I noticed this existing code to make the error output nicer:
> > https://github.com/seattleflu/backoffice/blob/70b67f7cee6ede75f28b109d0a459b1a0fa22d45/dev/refresh-database#L240-L243
> > 
> > From the [log you pasted in Slack](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1609808681024700?thread_ts=1609808492.024200&cid=C0173B4S1LY), it looks like that stopped working, maybe due to changes in error formatting?
> 
> I noticed that, too, and was surprised to see that the silencing of noisy logs no longer worked. I forgot to look into it, but I'll see if I can re-align that code with the new error formatting now.

I added four spaces to the beginning of the `Command was: ` filter in the perl command to reflect the latest Postgres error formatting.

kairstenfay:
I don't think this code is working as expected, yet. I'm currently trying to refresh the testing database from my local machine, and while the Postgres TOC errors are no longer appearing, the job is hanging silently.

tsibley (commented):
There's now two approaches in this code for suppressing ignorable errors:

1. Rewriting stdout/stderr by detecting when they occur.
2. Rewriting the TOC to avoid them from occurring in the first place

Approach 2 is cleaner in many ways, although does mean that the limitations of RDS instances get applied to local instances too, which is avoided by 1.

Notably, for comments on extensions, 1 is now redundant given 2.

We can't fully get away from 1 because of conditional role creation, but we could further reduce the scope of 1 by also filtering `TABLE DATA spatial_ref_sys` from the TOC.

dev/refresh-database
```diff
@@ -239,12 +241,12 @@ suppress-ignorable-errors() {
...
         s/(?-x:^pg_restore: .+? COMMENT EXTENSION (.+?)\s*)\n
           (?-x:^pg_restore: .+? ERROR:  must be owner of extension .+?)\n
-          (?-x:^Command was: COMMENT ON EXTENSION .+?)\n+
+          (?-x:^\s{4}Command was: COMMENT ON EXTENSION .+?)\n+
          /COMMENT EXTENSION \1 (failure due to RDS permissions ignored)\n/xmg;
 
         s/(?-x:^pg_restore: .+? TABLE DATA spatial_ref_sys rdsadmin)\n
           (?-x:^pg_restore: .+? ERROR:  permission denied for relation spatial_ref_sys)\n
-          (?-x:^Command was: COPY public\.spatial_ref_sys .+?)\n+
+          (?-x:^\s{4}Command was: COPY public\.spatial_ref_sys .+?)\n+
```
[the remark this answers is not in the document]
kairstenfay: Updated to use `\s*` in the latest version of the commit.

kairstenfay:
> There's now two approaches in this code for suppressing ignorable errors:
> 
>     1. Rewriting stdout/stderr by detecting when they occur.
> 
>     2. Rewriting the TOC to avoid them from occurring in the first place
> 
> 
> Approach 2 is cleaner in many ways, although does mean that the limitations of RDS instances get applied to local instances too, which is avoided by 1.
> 
> Notably, for comments on extensions, 1 is now redundant given 2.
> 
> We can't fully get away from 1 because of conditional role creation, but we could further reduce the scope of 1 by also filtering `TABLE DATA spatial_ref_sys` from the TOC.

Thanks for this feedback, @tsibley. I've also filtered `TABLE DATA spatial_ref_sys` in a new commit (that can likely be combined with the other `COMMENT - EXTENSION` commit) and removed the redundant regex logic to filter these errors from stdout/stderr. Please take a look and let me know if I have the right idea here.

dev/refresh-database
```diff
@@ -106,8 +106,11 @@ main() {
     log "Restoring filtered roles"
     psql --no-psqlrc < production-roles-filtered.sql |& suppress-ignorable-errors
 
+    log "Removing comments on extensions from the TOC"
+    pg_restore --list production.pgdb \
+        | grep -v 'COMMENT - EXTENSION|TABLE DATA spatial_ref_sys' > production.toc
```
tsibley, line 111: To use alternation (`|`) as-is you'll need to enable "extended regular expressions" with `-E`. The default is "basic regular expression", which treats `|` literally (but allows alternations via backslash escape as `\|`).

tsibley (changes requested):
Looks great! Thanks. One small detail in the TOC filtering, noted below.

dev/refresh-database
kairstenfay: Ah, thank you. I should have tested this change but forgot to because it runs for so long (on my machine, at least). 

I invoke the `-E` option in the latest comment.

tsibley:
I tested this locally and noticed the last commit wasn't working as intended. [Repushed](https://github.com/seattleflu/backoffice/compare/fd8375dff0cc92e94eaa0068076ad3f06c5d4ca1..c52aa3b7b67fe6aa6c7dfc47fe5dd84a6bab2030) to fix it.

kairstenfay:
> I tested this locally and noticed the last commit wasn't working as intended. [Repushed](https://github.com/seattleflu/backoffice/compare/fd8375dff0cc92e94eaa0068076ad3f06c5d4ca1..c52aa3b7b67fe6aa6c7dfc47fe5dd84a6bab2030) to fix it.

Thank you!

tsibley (approved):

PR #115, opened 2021-01-06
Automate linelist creation

kairstenfay:
Programatically create all of the linelists that we currently submit to
WA DoH:
    * SCAN (English, Russian, Vietnamese, Spanish, Chinese Traditional)
    * HCT
    * SCAN Kiosk
    * Shelters Asymptomatic
    * Shelters Intervention

The first step I took was to reproduce the exact results from Misja's
Jupyter Notebook. Then, I modified some columns by pulling in
standardized values from ID3C which is a better source for items like
collection date, encounter date, etc.

There are still some ways these scripts can be improved. For example,
in the `transform` job, it would probably be better in most cases to
export labels from REDCap rather than raw data that we have to create
meaningful labels for (e.g. see zipcode). Furthermore, the
`create-linelists` script could check for the existence of set
`REDCAP_API_TOKEN_redcap.iths.org_*` variables before exporting data
from ID3C. Or, it could accept a PROJECT_ID (optional) parameter it uses
to pass to the `transform` script.

But honestly, I've hit a point of diminishing returns on this project's
configuration. Additionally, we're likely to redo the custom format of
this linelist export into a new, standardized format (something like
ELFF), so I don't want to sink much more time into perfecting the
current format.

The fact that I've made Misja's life significantly easier (and Izzy's
and whoever's who eventually follows in their linelist footsteps) in
about one week makes me pretty happy. Note that the README instructions
are written for a tech novice who will be using this script for daily
work.

---------------------------

Depends on https://github.com/seattleflu/id3c-customizations/pull/177


kairstenfay:
[Automate linelist submission CSV export](https://trello.com/c/z7XDGtTz/952-automate-linelist-submission-csv-export)

bin/wa-doh-linelists/README.md
```diff
@@ -0,0 +1,91 @@
...
+See the specification for the linelist data format under `./bin/wa-doh-linelists/docs/SFS_DOH_Linelist_Data_Formats.xlsx`.
+
+The `./bin/wa-doh-linelists/create-linelists` script is the top-level script that exports ID3C data, merges it with REDCap report data, and performs transformations all in one.
+It requires test result `DATE` as an argument in `YYYY-MM-DD` format.
+It also requires the [REDCap API tokens environment variables](#redcap-tokens-environment-variables) and a [connection to an ID3C database](#connecting-to-the-production-id3c-database) (see below).
+
+Once you've configured your environment properly, run the following to create linelists:
+```sh
+./bin/wa-doh-linelists/create-linelists 2020-12-28
+```
```
kairstenfay, line 13: TODO: describe the `--output-dir` default and data naming scheme

bin/wa-doh-linelists/create-linelists
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
jccraft-uw, line 1: To match our other "pipeline" projects under backoffice/bin could you name this file 'generate' instead of 'create-linelists'?

bin/wa-doh-linelists/transform
```diff
@@ -0,0 +1,234 @@
...
+
+logging.captureWarnings(True)
+
+log = logging.getLogger(__name__)
+log.setLevel(LOG_LEVEL)
+
+base_dir = Path(__file__).resolve().parent.parent.parent.resolve()
+
+
+def export_redcap_report(config: Dict[str, Any]) -> Dict[str, Any]:
```
jccraft-uw, line 39: This would be a really useful function to have in ID3C in lib/id3c/cli/redcap.py. I can see us needing to export a REDCap report in the future.

bin/wa-doh-linelists/transform
```diff
@@ -0,0 +1,234 @@
...
+    return redcap_report
+
+def fill_missing_data(redcap_report: pd.DataFrame, config: Dict[str, Any]) -> pd.DataFrame:
+    """ """
+    # Convert empty strings to NA. We have to do this step after we relabel
+    # values, unless we want to deal with mixed dtypes of object and string.
+    redcap_report.replace({'': pd.NA}, inplace=True)
+
+    # Backfill missing data in some columns with a fallback value
+    if config.get('column_backfill_map'):
```
jccraft-uw, line 93: Oh wow. It's interesting to see how much data value manipulation is required.

bin/wa-doh-linelists/transform
```diff
@@ -0,0 +1,234 @@
...
+    data[date_columns] = \
+        data[date_columns] \
+            .apply(pd.to_datetime, errors='coerce') \
+            .apply(lambda date: date.dt.strftime('%m/%d/%Y'))
+
+    return data
+
+def overwrite_with_phi(linelist: pd.DataFrame, config: Dict[str, Any]) -> pd.DataFrame:
+    """
+    Overwrite certain ID3C data in a given *linelist* with additional PHI we
```
jccraft-uw, line 123: Couldn't we just not query these fields from ID3C and instead join the REDCap export data to pull them in?

bin/wa-doh-linelists/transform
```diff
@@ -0,0 +1,234 @@
...
+
+        # Load REDCap report data. Add columns with static values, strip
+        # strings of commas, and convert empty strings to NA
+        redcap_report = pd.DataFrame.from_dict(
+            export_redcap_report(config), dtype='string') \
+            .rename(columns=config['column_rename_map']) \
+            .assign(**config['static_columns']) \
+            .apply(lambda c: c.str.replace(',', ''))
+
+        if config.get('columns_to_drop'):
```
jccraft-uw, line 203: Would it make things a lot cleaner for us if a REDCap report specifically for the linelist was created in each REDCap project so we wouldn't have to jump through this kind of hoop? I'm not saying that we should make those, but that could be someone we push for.

bin/wa-doh-linelists/transform
kairstenfay: There are only a couple of projects where we overwrite ID3C columns with REDCap data: HCT and Shelters.  I also realize now that this docstring is misnamed. In many cases we're overwriting REDCap data with different REDCap data. The only ID3C columns we're overwriting in this method are sex, ethnicity, and site name.

bin/wa-doh-linelists/transform
kairstenfay: I'm not sure I understand your question. This project specifically uses REDCap reports created for linelists. Are you suggesting we ask Misja to delete columns from the REDCap report that match our ID3C export identically?

kairstenfay: I followed up with Misja and she is OK with deleting the `columns_to_drop` columns from the various REDCap reports for the relevant projects. So, this section can go away, but we'll want to coordinate deployment with her column deletion.

jccraft-uw: Cool. Thanks for talking about this in standup. Since these reports are specifically for the linelists we can remove the columns from the reports instead of doing that inside the job.

bin/wa-doh-linelists/create-linelists
kairstenfay: Renamed

bin/wa-doh-linelists/transform
kairstenfay: I made a note at the top of this PR to coordinate dropping the specified columns from REDCap reports. At that point, I'll also drop this part of the code. I don't want to drop it prematurely, because the output linelist CSVs will be broken until those columns are removed.

bin/wa-doh-linelists/transform
kairstenfay: I've updated the docstring and function name.

bin/wa-doh-linelists/README.md
```diff
@@ -0,0 +1,106 @@
...
+## Deep dive
+The `generate` script invoked above is comprised of two, separate scripts.
+Calling these individual scripts is useful if you want lower level control over the input and output of the generated linelists.
+
+1. One of them exports linelist data from ID3C given a date (and requires a connection to an ID3C database):
+    ```sh
+    ./bin/wa-doh-linelists/export-id3c-hcov19-results 2020-12-28 > ./bin/wa-doh-linelists/data/id3c-export-2020-12-28.csv
+    ```
+2. The other merges ID3C linelist data and REDCap report data, transforming and standardizing them into a format WA DoH expects.
+   This script reqiures a test result `--date` in YYYY-MM-DD format and `--id3c-data` which, if you choose to run this script instead of `generate`, you'll need to export using the `export-id3c-hcov19-results` script explaind above.
```
kschwabefry, line 27: nit typo: explained

bin/wa-doh-linelists/README.md
```diff
@@ -0,0 +1,106 @@
...
+[seattleflu-production]
+host=production.db.seattleflu.org
+user=my-user-name
+dbname=production
+port=5432
+```
+
+I may also choose to store my password inside a separate file, `~/.pgpass`, using this as an example:
+```
+production.db.seattleflu.org:5432:*:my-user-name:my-pasword
```
kschwabefry, line 62: nit: typo password

bin/wa-doh-linelists/README.md
```diff
@@ -0,0 +1,106 @@
...
+ls
+```
+
+There, you'll see several folders.
+We'll be targeting two of them for creating linelists.
+You'll need to copy your REDCap API token from each project and create a new file containing only the token.
+Here's an example of creating a new environment variable for the SCAN IRB English project with the text editor `nano`:
+```sh
+nano redcap-scan/REDCAP_API_TOKEN_redcap.iths.org_22461
+```
```
kschwabefry, line 90: Suggest adding "(project ID 22461)" to clarify for reader the significance of that id in the example, like:

> Here's an example of creating a new environment variable for the SCAN IRB English project **(project ID 22461)** with the text editor

bin/wa-doh-linelists/transform
```diff
@@ -0,0 +1,234 @@
...
+import pandas as pd
+from sys import stderr
+from pathlib import Path
+from typing import Any, Dict
+from datetime import datetime
+
+# Disable chained assignments
+pd.options.mode.chained_assignment = None
+
+LOG_LEVEL = os.environ.get("LOG_LEVEL", "debug").upper()
```
kschwabefry, line 23: I wonder how much debugging output this process generates. May be useful to default this to a higher log level like warning/error for users running this?

etc/wa-doh-linelists.yaml
```diff
@@ -0,0 +1,675 @@
...
+  "uw_greek_house",
+  "priority_code",
+  "site_name",
+  "site_context",
+  "participant_role",
+  "collection_barcode"
+]
+raw_values_map: {
+  home_zipcode_2: {
+    '44': '98101',
```
kschwabefry, line 139: This is a lot of config to maintain! Is there a way to pull these mappings from shared code in id3c?

kschwabefry (approved):
This is really exciting to see automated, I know it will save a lot of time! 
I left a few comments, but nothing I view as a blocker.
I didn't download this and try it out, but for the sake of thoroughness, I did want to be sure to explicitly ask about the testing story for the new process. Have you or Misja gone through and compared output from current vs new process?

kairstenfay:
> This is really exciting to see automated, I know it will save a lot of time!
> I left a few comments, but nothing I view as a blocker.
> I didn't download this and try it out, but for the sake of thoroughness, I did want to be sure to explicitly ask about the testing story for the new process. Have you or Misja gone through and compared output from current vs new process?

Thanks for your feedback! Yes, I have reproduced the linelists from 12-28-2020 (and 12-27-2020 for asymptomatic shelters) identically* with this script (compared to Misja's "source of truth"). 

* this script actually zero-pads months and days < 10, and pulls in extra dates (like collection date) that REDCap was missing for a few rows

etc/wa-doh-linelists.yaml
kairstenfay: Ah yep, good point. I forgot we accounted for this in id3c-customization's `redcap_det_scan`

bin/wa-doh-linelists/README.md
kairstenfay: I have added this detail to the instructions.

bin/wa-doh-linelists/transform
kairstenfay: I have lowered the log level to `info`.

etc/wa-doh-linelists.yaml
kairstenfay: I am now using the redcap-det scan logic from id3c-customizations to map zipcode labels instead of duplicating that effort here.

bin/wa-doh-linelists/transform
kairstenfay: I mentioned this in stand-up this morning, but upon further consideration, I'm a bit worried about dropping columns from the REDCap report, thereby destroying the "old way" of doing things. 

I can foresee certain situations where someone like Misja or Izzy is in a time crunch, unable to connect to ID3C at the command line (as in with Misja's firewall issues this morning), or otherwise (understandably) skeptical of our pipeline results and instead waning to try the "old way" (using their original Metabase export + REDCap report export + Jupyter notebook workflow).

I still think it's better to pull from ID3C in every place possible for multiple reasons.
1. ID3C data are standardized. Right now field labels for "sex" are consistently capitalized, etc. in our REDCap projects, but this won't necessarily always be true.
2. By pulling from ID3C, we don't need to map raw values for e.g. sex to their human readable values. That is the default value in ID3C. If we stop pulling `sex` from ID3C, then I have to re-introduce the `raw_values_map` configuration for SCAN and HCT, which puts us in a trickier spot than just dropping `sex_new` from REDCap, IMO. 
3. ID3C is able to "find" data missing in REDCap. For example, my script brings in additional data for `collection_date` (likely from the manifest) that are absent in REDCap for whatever reason. 

The only other solution to avoiding the `columns_to_drop` config entry and related code (but preserving the original columns in the REDCap report) is to explicitly mention the columns from REDCap that we want to keep. This ends up being much more verbose.

After all considerations, I don't see a better current solution than preserving the `columns_to_drop` config and code while also preserving the conflicting REDCap report columns.

kairstenfay: I just confirmed with Misja that she wants to preserve the old way of creating linelists, therefore, we won't delete the REDCap report columns are named identically to the `shipping.linelist_data_for_wa_doh_v1` columns.

kairstenfay:
Misja just informed me that she's currently using these scripts to submit data to WA DoH, and thus, the code in this PR is considered production code. I'm going to go ahead and merge for now, but I'm open to more feedback, so please feel free to continue to comment on this PR.

PR #120, opened 2021-01-25
Cron job adjustments

jccraft-uw:
Please see notes in the commit messages. 
1. Use flock for refresh materialized view jobs.
2. Run the refresh materialized view jobs less frequently.
3. Reduce the UW Reopening ETL's redcap-api-batch-size.

Before deploying this, create a flock directory on the backoffice server:
/home/ubuntu/flock .

### DIFF
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -120,14 +120,14 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 * * * * * ubuntu promjob "terminate-idle-sessions" fatigue --quiet terminate-idle-sessions
 # Run the old metabase session disconnector routine every minute
 * * * * * ubuntu promjob "terminate-old-metabase-sessions" fatigue --quiet terminate-old-metabase-sessions
-# Run the refresh materialized view routine every 5 minutes include latest
+# Run the refresh materialized view routine every 10 minutes include latest
 # ingested FHIR documents in the view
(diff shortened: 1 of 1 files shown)


tsibley (approved):
This will use lockfiles in `/home/ubuntu` (the cronjobs' working dir) named after the views, right?

In general I'd like to be a little more organized there, but that seems fine for now. I think the better time spent over manually organizing this would be moving to systemd timers.

jccraft-uw:
> This will use lockfiles in `/home/ubuntu` (the cronjobs' working dir) named after the views, right?
> 
> In general I'd like to be a little more organized there, but that seems fine for now. I think the better time spent over manually organizing this would be moving to systemd timers.

How about this change?
I'm now putting the flock files explicitly under /home/ubuntu/flock
Note that the existing jobs that use flock use the $GEOCODING_CACHE file which is in /home/ubuntu.

tsibley:
> How about this change?
> I'm now putting the flock files explicitly under /home/ubuntu/flock

I like pulling out the location into `FLOCK_DIR`. I'll note that the conventional location for lock files is under `/var/run/lock`, which we do use for return-of-results.

jccraft-uw:
> > How about this change?
> > I'm now putting the flock files explicitly under /home/ubuntu/flock
> 
> I like pulling out the location into `FLOCK_DIR`. I'll note that the conventional location for lock files is under `/var/run/lock`, which we do use for return-of-results.

The `var` path looks much better, so I switched to that.

jccraft-uw:
Deployed

PR #122, opened 2021-01-26
Linelist formatting

kairstenfay:
This PR addresses the formatting issues Misja recently raised in Slack regarding the new linelists we submit to WA DoH.

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -246,4 +246,5 @@ if __name__ == "__main__":
 
     mega_linelist.sort_values(by=['record_id']) \
         .to_csv(args.output_dir / f"linelist_{date.replace('/', '')}.csv",
-            index=False)
+            index=False,
+            line_terminator='\r\n')
--- a/etc/wa-doh-linelists.yaml
+++ b/etc/wa-doh-linelists.yaml
@@ -132,7 +132,6 @@ column_names: [
   "site_name",
   "site_context",
   "participant_role",
-  "collection_barcode"
 ]
 
 ---


jccraft-uw (approved):
It's neat that .to_csv takes the line_terminator.

I think the commit (wa-doh-linelists: Use macOS CRLFs) should say Windows instead of macOS.

kairstenfay:
> It's neat that .to_csv takes the line_terminator.
> 
> I think the commit (wa-doh-linelists: Use macOS CRLFs) should say Windows instead of macOS.

Oh, I didn't do my due diligence to actually figure out which OS uses `\r\n`. So, if Misja is creating CSVs with a Mac, why did the file end up with `\r\n` CRLFs? Is that because she was using Excel (a Microsoft product)?

jccraft-uw:
> > It's neat that .to_csv takes the line_terminator.
> > I think the commit (wa-doh-linelists: Use macOS CRLFs) should say Windows instead of macOS.
> 
> Oh, I didn't do my due diligence to actually figure out which OS uses `\r\n`. So, if Misja is creating CSVs with a Mac, why did the file end up with `\r\n` CRLFs? Is that because she was using Excel (a Microsoft product)?

That's what I gathered, that it's Excel that's doing the '\r' part. From https://en.wikipedia.org/wiki/Newline#History it looks like MacOS follows Unix/Linux.
But it's just a commit message. It doesn't have to be the gospel!

tsibley:
While the resulting reasoning seems correct, based on the vocabulary used in this PR and Slack, I think there's some confusion/misconceptions around line endings. I'll try to briefly clear them up.

```
Carriage return = CR = \r = byte 0x0d in ASCII and UTF-8
Line feed       = LF = \n = byte 0x0a in ASCII and UTF-8
```

A CRLF is not a special character that changes morphs based on the OS. What changes based on the OS/application is the character/bytes used for line endings in plain text.

Unixes (including macOS/OS X) and Linux conventionally use just LFs.

OG Macintoshes conventionally used just CRs.

Windows conventionally uses CRLFs (that is, two bytes instead of one).

Many programs will default to the OS convention, but also let you choose which to use.  Excel and MS Office applications have in the past (and maybe still it seems) default to Windows line endings even when not running on Windows, to be more compatible with their Windows versions, but they also will let you choose line endings when saving.

Some textual network protocols and file formats that are fundamentally cross-platform require a particular line ending so data isn't ambiguous (typically LF, but sometimes CRLF). Others try to do the right thing transparently (with varying levels of success).

kairstenfay:
> While the resulting reasoning seems correct, based on the vocabulary used in this PR and Slack, I think there's some confusion/misconceptions around line endings. I'll try to briefly clear them up.
> 
> ```
> Carriage return = CR = \r = byte 0x0d in ASCII and UTF-8
> Line feed       = LF = \n = byte 0x0a in ASCII and UTF-8
> ```
> 
> A CRLF is not a special character that morphs based on the OS. What changes based on the OS/application are the characters/bytes used for line endings in plain text.
> 
> Unixes (including macOS/OS X) and Linux conventionally use just LFs.
> 
> OG Macintoshes conventionally used just CRs.
> 
> Windows conventionally uses CRLFs (that is, two bytes instead of one).
> 
> Many programs will default to the OS convention, but also let you choose which to use. Excel and MS Office applications have in the past (and maybe still it seems) default to Windows line endings even when not running on Windows, to be more compatible with their Windows versions, but they also will let you choose line endings when saving.
> 
> Some textual network protocols and file formats that are fundamentally cross-platform require a particular line ending so data isn't ambiguous (typically LF, but sometimes CRLF). Others try to do the right thing transparently (with varying levels of success).

Thanks so much for this clarification, Tom. I just re-pushed my commit message to rephrase the reason for the changes in a hopefully more accurate way.

PR #123, opened 2021-01-27
uw-ehs-report: Use NDJSON instead of CSV for the intermediate file format

tsibley:
Motivated by the problems with the UW EH&S report yesterday.

See commit messages for details.

Todo:

- [x] Merge and push https://github.com/seattleflu/id3c/pull/206
- [ ] `pipenv update` in _id3c-production/_ and commit to this branch before merging

### DIFF
--- a/bin/uw-ehs-report/generate
+++ b/bin/uw-ehs-report/generate
@@ -26,16 +26,7 @@ main() {
         esac
     done
 
-    datadir="$(TMPDIR="$base" mktemp -d -t uw_ehs_report-XXXXXX)"
-
-    # Always leave and delete this temporary directory regardless of exit code
-    trap "rm -rf '$datadir'" EXIT
-
-    # Generate the report NDJSON
-    "$bin"/generate-report > "$datadir/uw_ehs_report.ndjson"
-
-    # Import the NDJSON into the transfer REDCap project.
-    "$bin"/import-to-uw-ehs-transfer-redcap "$datadir/uw_ehs_report.ndjson"
+    "$bin"/generate-report | "$bin"/import-to-uw-ehs-transfer-redcap -
 }
 
 print-help() {
--- a/bin/uw-ehs-report/generate
+++ b/bin/uw-ehs-report/generate
@@ -31,11 +31,11 @@ main() {
     # Always leave and delete this temporary directory regardless of exit code
     trap "rm -rf '$datadir'" EXIT
 
-    # Generate the report CSV
-    "$bin"/generate-report-csv > "$datadir/uw_ehs_report.csv"
+    # Generate the report NDJSON
+    "$bin"/generate-report > "$datadir/uw_ehs_report.ndjson"
 
-    # Import the CSV into the transfer REDCap project.
-    "$bin"/import-to-uw-ehs-transfer-redcap "$datadir/uw_ehs_report.csv"
+    # Import the NDJSON into the transfer REDCap project.
+    "$bin"/import-to-uw-ehs-transfer-redcap "$datadir/uw_ehs_report.ndjson"
 }
 
 print-help() {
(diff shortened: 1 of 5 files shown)


bin/uw-ehs-report/import-to-uw-ehs-transfer-redcap
```diff
@@ -32,20 +30,17 @@ if __name__ == '__main__':
...
+        type=argparse.FileType("r"),
+        help="An NDJSON file containing the data to import into EHS REDCap project")
 
     args = parser.parse_args()
 
     project = redcap.Project("https://redcap.iths.org", 24025)
 
-    ehs_dataset = pd.read_csv(args.filename)
+    # Post data to REDCap in batches of 8,500 records to reduce strain on REDCap servers
+    redcap_post_batch_size = 8500
```
kschwabefry, line 43: The hard part about tuning this is that this is a moving target. I think the size of each record matters in addition to the number of records. We may have to lower this in time, but 8.5k works ok currently.

bin/uw-ehs-report/generate
```diff
@@ -26,16 +26,7 @@ main() {
...
-
-    # Always leave and delete this temporary directory regardless of exit code
-    trap "rm -rf '$datadir'" EXIT
-
-    # Generate the report CSV
-    "$bin"/generate-report-csv > "$datadir/uw_ehs_report.csv"
-
-    # Import the CSV into the transfer REDCap project.
-    "$bin"/import-to-uw-ehs-transfer-redcap "$datadir/uw_ehs_report.csv"
+    "$bin"/generate-report | "$bin"/import-to-uw-ehs-transfer-redcap -
```
kschwabefry, line 29: I like the idea of piping this in here. It makes me wonder if we would want to do something similar in the return of results pipeline, since that also writes to disk temporarily, but in that case we use that results data set both for PDF generation and as an actual file to upload into S3. So, that might be more complicated to change.

kschwabefry (approved):

bin/uw-ehs-report/generate
jccraft-uw: What does the final "-" do here at the end of the line?

kairstenfay: The `-` is a special character in bash that allows you to read from stdin or write to stdout without specifying a file

tsibley: `-` is a long-standing Unix convention for commands that take filenames that means "read from stdin" or "write to stdout". However, it is not specially-handled by the shell; the program sees the literal `-` as its argument and must explicitly handle it (e.g. try `ls -`). In this case, [`argparse.FileType`](https://docs.python.org/3/library/argparse.html#filetype-objects) does the conventional thing for us. In ID3C, we use [`click.File`](https://click.palletsprojects.com/en/7.x/api/#click.File) similarly.

bin/uw-ehs-report/import-to-uw-ehs-transfer-redcap
tsibley: Yep, though note that tuning wasn't the reason I changed it: I changed it so it matched the reality of what the code's been doing ([details](https://github.com/seattleflu/backoffice/pull/123/commits/d7dd9bfa4b2187f248490d2871ad01236ff76ad8)).

bin/uw-ehs-report/generate
kairstenfay: Neat, thanks for that explanation, Tom.

jccraft-uw: Thanks a lot for that info, @kairstenfay and @tsibley . I was trying to understand how "-" matches up with the input argument of import-to-uw-ehs-transfer-redcap . That's pretty neat.

jccraft-uw (approved):

tsibley:
Repushed to rebase onto latest master and [upgrade ID3C](https://github.com/seattleflu/backoffice/compare/ee565a4b476f19b2af5f4b12111e769dfad675dc..c82f3975cc18f659e4032c5176e2eef5d3311efb#diff-1d2f2477f3e2924eb4286939b8c7eaf59fbe13943e06e24974ba31c9bdcaddb2) for the `dump_ndjson()` improvement.

tsibley:
Deployed.

PR #124, opened 2021-01-28
Linelist merge error

kairstenfay:
Exit with a helpful message when duplicated barcodes are found in REDCap when creating the WA DOH Linelist export.

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -210,11 +210,28 @@ if __name__ == "__main__":
         redcap_report = redcap_report[redcap_report.collection_barcode.notnull()]
         log.debug(f"Dropped {prev - len(redcap_report)} row(s) with a missing `collection_barcode`")
 
-        # Merge REDCap and ID3C data
-        linelist = id3c_data.merge(redcap_report,
-            how='inner',
-            on='collection_barcode',
-            validate='1:1')
+        try:
+            # Merge REDCap and ID3C data
+            linelist = id3c_data.merge(redcap_report,
+                how='inner',
+                on='collection_barcode',
+                validate='1:1')
+
+        # Abort if duplicated collection barcodes exist. We don't want this
+        # pipeline to choose which is the "right" barcode for a record.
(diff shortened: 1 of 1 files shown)


bin/wa-doh-linelists/transform
```diff
@@ -228,11 +228,25 @@ if __name__ == "__main__":
...
+        except pd.errors.MergeError:
+            mask = redcap_report.duplicated(subset='collection_barcode', keep=False)
+            duplicated_collection_barcodes = redcap_report[mask] \
+                [['record_id', 'collection_barcode']] \
+                .sort_values(by='collection_barcode') \
+                .reset_index(drop=True)
+
+            raise Exception("Duplicate barcodes detected in project "
+                f"{config['project_id']!r}: {config['name']!r}\n\n"
+                f"{duplicated_collection_barcodes}")
```
kschwabefry, line 249: Adding a more detailed and user-friendly exception is a great idea. I think recasting all MergeErrors may be too general here, as it can be caused by other situations beyond duplicated barcodes/non-unique merge key. If there aren't duplicates, the exception will still be thrown, with an empty dataframe for duplicated_collection_barcodes. In the case, we'd want to keep the details from the original merge error.

How about:

```
if duplicated_collection_barcodes.empty:
    raise #Re-raise original MergeException
else:
    raise Exception("Duplicate barcodes detected in project "
                f"{config['project_id']!r}: {config['name']!r}\n\n"
                f"{duplicated_collection_barcodes}")
```

kschwabefry (changes requested):

bin/wa-doh-linelists/transform
kairstenfay: Great point. I had not considered that MergeErrors exist outside of the `validation` assertion failure. 

I just re-pushed following your suggestion.

kschwabefry (approved):

kairstenfay:
[Linelists: Better merge error handling](https://trello.com/c/CCuh1Di1/1017-linelists-better-merge-error-handling)

PR #126, opened 2021-01-29
Stop silencing errors from the presence-absence ETL

jccraft-uw:
We went the entire day without noticing that the presence-absence ETL
had been failing since last night. In this case I would rather see
Slack alert entries every 15 minutes than to go a day without seeing
a problem with our results release pipeline.


tsibley (approved):

tsibley:
For posterity, there's some more convo about this [in Slack](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1611888871116800).

jccraft-uw:
Deployed.

PR #127, opened 2021-02-02
Accommodate afh

jccraft-uw:
For the AFH/Workplace project we needed a new code path to handle "classic" REDCap projects that don't follow all the rules of SCAN (doesn't require contacting the participant before releasing a positive or inconclusive result). I used the idea done with the longitudinal path to create a config-driven classic path.

When generating PDFs, we now generate a PDF in all languages for "IRB" results and use a different template for English "clinical" results.

This depends on the corresponding template changes in https://github.com/seattleflu/lab-result-reports/pull/28/files

### DIFF
--- a/bin/return-of-results/export-redcap-sfs-classic
+++ b/bin/return-of-results/export-redcap-sfs-classic
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# usage: export-redcap-sfs-classic
+#
+# Exports participant information from SFS "classic" mode REDCap projects as NDJSON.
+# REDCap projects are processed according to the YAML configuration in $CONFIG,
+# which defaults to etc/sfs-classic-redcap-projects.yaml.
+#
+# The $CONFIG file is expected to have at least one YAML document in it with
+# the following format:
+#
+# ---
+# redcap:
+#   url: https://redcap.iths.org
+#   project: 27619 # Clinical COVID Testing (Congregate Settings)
+# source: adult_family_home_workplace_outbreak
+# contact_before_releasing_result: false
+# contacted_field: null
+# participant_first_name_field: core_participant_first_name
+# participant_last_name_field: core_participant_last_name
+# participant_birthdate_field: core_birthdate
+# additional_fields:
+#   - {"name":"ordering_provider", "required": true}
+# prioritized_barcode_fields:
+#   - core_collection_barcode
+#   - return_collection_barcode
+
+import sys
+import json
+import yaml
+from pathlib import Path
+from id3c.cli.redcap import Project
+from os import environ
+from typing import Dict, List, Any
+
+config_file  = environ.get("CONFIG")
+
+if not config_file:
+    base_dir = Path(__file__).resolve().parent.parent.parent
+    config_file = base_dir / "etc/sfs-classic-redcap-projects.yaml"
+
+
+class SFSProject(Project):
+    source: str
+    contact_before_releasing_result: bool
+    contacted_field: str
+    participant_first_name_field: str
+    participant_last_name_field: str
+    participant_birthdate_field: str
+    additional_fields: List[Dict[str, Any]]
+    prioritized_barcode_fields: List[str]
+
+    def __init__(self, config: Dict[str, Any]):
+        super().__init__(config['redcap']['url'], config['redcap']['project'])
+        self.source = config['source']
+        self.contact_before_releasing_result = config['contact_before_releasing_result']
+        self.contacted_field = config.get('contacted_field')
+        self.participant_first_name_field = config['participant_first_name_field']
+        self.participant_last_name_field = config['participant_last_name_field']
+        self.participant_birthdate_field = config['participant_birthdate_field']
+        self.additional_fields = config.get('additional_fields') or []
+        self.prioritized_barcode_fields = config['prioritized_barcode_fields']
+
+
+def main():
+    with open(config_file, 'r') as f:
+        configs = list(yaml.safe_load_all(f))
+
+    for config in configs:
+        project = SFSProject(config)
+
+        assert (not project.contact_before_releasing_result or (project.contact_before_releasing_result and project.contacted_field)), \
+        "If the project requires contact before releasing the result, you must specify the `contacted_field` field. Correct the project " \
+        f"with source: {project.source}."
+
+        for record in fetch_records(project):
+            record['source'] = project.source
+            record['contact_before_releasing_result'] = str(project.contact_before_releasing_result).lower()
+
+            barcode = None
+
+            for field in project.prioritized_barcode_fields:
+                if record.get(field):
+                    barcode = record[field]
+                    break
+
+            record['prioritized_barcode'] = barcode
+
+            # Use the record only if it has the necessary barcode fields populated
+            if barcode:
+
+                # Don't retain the raw barcode fields now that we have the prioritized one
+                for field in project.prioritized_barcode_fields:
+                    del record[field]
+
+                print(json.dumps(record, indent = None, separators = ",:"), flush = True)
+
+
+def fetch_records(project) -> List[Dict[str, dict]]:
+    """
+    Fetch all records from the REDCap project.
+    """
+    redcap_records = []
+
+    participant_fields = {project.participant_first_name_field,
+        project.participant_last_name_field, project.participant_birthdate_field}
+
+    fields = {project.record_id_field, *participant_fields, *project.prioritized_barcode_fields,
+        *[entry['name'] for entry in project.additional_fields]}
+
+    required_fields = {*participant_fields,
+        *[entry['name'] for entry in project.additional_fields if entry['required'] == True]
+        }
+
+    if project.contacted_field:
+        fields.add(project.contacted_field)
+        required_fields.add(project.contacted_field)
+
+    for record in project.records(fields = fields, raw = True):
+        if not all(len(record[field]) for field in required_fields):
+            continue
+
+        # Standardize the name of some keys
+        record['core_participant_first_name'] = record.pop(project.participant_first_name_field)
+        record['core_participant_last_name'] = record.pop(project.participant_last_name_field)
+        record['core_participant_birth_date'] = record.pop(project.participant_birthdate_field)
+
+        if project.contacted_field:
+            record['participant_contacted'] = record.pop(project.contacted_field)
+        else:
+            # Use "" to match a REDCap unset variable
+            record['participant_contacted'] = ""
+
+        redcap_records.append(record)
+
+    return redcap_records
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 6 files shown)


jccraft-uw:
[Include AFH studies in RoR pipeline](https://trello.com/c/IKJlgQ3k/969-include-afh-studies-in-ror-pipeline)

bin/return-of-results/transform
```diff
@@ -182,7 +212,10 @@ def edit_status_code(results: pd.DataFrame) -> pd.DataFrame:
     is_inconclusive = results['status_code'] == 'inconclusive'
 
     participant_contacted = results['contacted'] == 'yes'
-    require_contacted = results['source'].isin(['scan', 'childcare'])
+
+    require_contacted_lambda = lambda row: (row['source'] is not pd.NA and row['source'] in ['scan', 'childcare']) \
+        or (row['contact_before_releasing_result'] is not pd.NA and row['contact_before_releasing_result'] == 'true')
```
kschwabefry, line 217: Why do we have the check for `row['contact_before_releasing_result'] is not pd.NA`?

bin/return-of-results/export-redcap-sfs-classic
```diff
@@ -0,0 +1,139 @@
...
+        required_fields.add(project.contacted_field)
+
+    for record in project.records(fields = fields, raw = True):
+        if not all(len(record[field]) for field in required_fields):
+            continue
+
+        # Standardize the name of some keys
+        record['participant_first_name'] = record.pop(project.participant_first_name_field)
+        record['participant_last_name'] = record.pop(project.participant_last_name_field)
+        record['participant_birth_date'] = record.pop(project.participant_birthdate_field)
```
kschwabefry, line 125: I think core_participant_* (e.g. core_participant_first_name) are the standardized names - we use them in the longitudinal projects as well and I think intend to push them going forward. If we are going to rename fields in the record here during processing, I think it would be more consistent to standardize to core_participant_* instead of just participant_*

etc/sfs-classic-redcap-projects.yaml
```diff
@@ -0,0 +1,15 @@
+---
+redcap:
+  url: https://redcap.iths.org
+  project: 27619 # Clinical COVID Testing (Congregate Settings)
+source: adult_family_home_workplace_outbreak
+contact_before_releasing_result: false
+contacted_field: null
+participant_first_name_field: core_participant_first_name
+participant_last_name_field: core_participant_last_name
+participant_birthdate_field: core_birthdate
```
kschwabefry, line 10: (non-blocker) this could be more structured if done as a dict, like:

```
project_fields:
  participant_contacted: null
  participant_first_name: core_participant_first_name
  participant_last_name: core_participant_last_name
  participant_birthdate: core_birthdate
```

They would be accessed in `export-redcap-sfs-classic` like `config['project_fields']['participant_contacted']`, `config['project_fields']['participant_first_name']`

bin/return-of-results/generate-pdfs
```diff
@@ -8,7 +8,7 @@ main() {
     local output="${2:-s3://dokku-stack-phi/covid19/results-scan-study/{qrcode\}-{birth_date\}}"
     local -a mounts
     mounts=()
-    local docker_container_build_number="27"
+    local docker_container_build_number="28"
```
kschwabefry, line 11: May be useful to link to https://github.com/seattleflu/lab-result-reports/pull/28/files in the PR comments as a prerequisite for merging this.

bin/return-of-results/transform
```diff
@@ -221,4 +221,4 @@ if __name__ == '__main__':
     joined_data = redcap_data.merge(id3c_data, how='inner', on='qrcode')
     joined_data = edit_status_code(joined_data)
 
-    joined_data[['qrcode', 'pat_name', 'birth_date', 'collect_ts', 'result_ts', 'status_code', 'swab_type', 'staff_observed']].to_csv(args.output, index=False)
+    joined_data[['qrcode', 'pat_name', 'birth_date', 'collect_ts', 'result_ts', 'status_code', 'swab_type', 'staff_observed', 'pre_analytical_specimen_collection']].to_csv(args.output, index=False)
```
kschwabefry: I like the note in the commit about merging with caution, that's a great call out. I think its worth adding a comment here into the code as well to the same effect, for our future selves. Something like:

`Caution: a side effect of adding columns here is that the pipeline will consider all results as new when it diffs against the previous run's output, and attempt to regenerate all results PDFs. That process is incredibly slow, so it may be prudent to manually backfill new columns into the previous_results.csv in S3`

When we deploy this, what if we disable the RoR cron job, so that it doesn't trigger while the CSVs used to diff are being manually manipulated to backfill the new columns? Disabling the cron job could be another commit bundled with this PR as a safeguard

bin/return-of-results/transform
```diff
@@ -241,7 +241,10 @@ if __name__ == '__main__':
     sfs_longitudinal_redcap_data = parse_sfs_longitudinal_redcap(args.sfs_longitudinal_redcap_data)
     sfs_classic_redcap_data = parse_sfs_classic_redcap(args.sfs_classic_redcap_data)
 
-    redcap_data = scan_redcap_data.append(sfs_longitudinal_redcap_data).append(sfs_classic_redcap_data)
+    # Combine the DataFrames by appending to the first.
+    # Use `ignore_index = True` to prevent duplicate index values.
+    redcap_data = scan_redcap_data.append(sfs_longitudinal_redcap_data, ignore_index = True) \
+        .append(sfs_classic_redcap_data, ignore_index = True)
```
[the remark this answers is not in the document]
jccraft-uw: The error is not in the actual appending but later when trying to update the appended data. The error came here:
` # Check for duplicate barcodes across ALL projects.
    redcap_data["qrcode"] = drop_duplicate_barcodes(redcap_data["qrcode"])`

In redcap_data there were duplicate index values, so trying to update the qrcode series failed.

bin/return-of-results/transform
jccraft-uw: The contact_before_releasing_result field exists in only one of the 3 DataFrames that get appended together, so data from the other 2 DataFrames have contact_before_releasing_result = pd.NA. 
Without this check for NA, the value of require_contacted would be pd.NA for a row where contact_before_releasing_result == pd.NA.
That caused the following statement to fail because it was trying to compare pd.NA to a boolean.
results.loc[(is_positive | is_inconclusive) & ~(participant_contacted) & require_contacted, 'status_code'] = 'pending'

bin/return-of-results/generate-pdfs
jccraft-uw: Done

bin/return-of-results/export-redcap-sfs-classic
jccraft-uw: Might as well. I changed those to 'core_'.

bin/return-of-results/transform
jccraft-uw: I added your caution comment in.

I've gotten good at manually disabling cron jobs in prod while I tinker. If there's risk that someone else will deploy while I'm updating the file in S3 then I'll do that as a PR.

etc/sfs-classic-redcap-projects.yaml
jccraft-uw: I see what you mean, but it doesn't seem like we need to add another layer because all of the settings are project fields, even including prioritized_barcode_fields.

bin/return-of-results/generate-pdfs
```diff
@@ -52,13 +53,31 @@ main() {
...
+            --env=AWS_SECRET_ACCESS_KEY \
+            --env=LOG_LEVEL \
+            --user "$(id -u):$(id -g)" \
+            "${mounts[@]}" \
+            seattleflu/lab-result-reports:build-$docker_container_build_number \
+                fill-template \
+                    --template "clinical/report-en.tex" \
+                    --params "$params" \
+                    --output "$output-en.pdf" \
+                    --filter 'status_code not in ["not-received", "pending"] and pre_analytical_specimen_collection == "clinical" '
```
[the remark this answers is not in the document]
jccraft-uw: That's a really good point. The only place where we access the data exported from ID3C is in the `transform' script. I could put the check right after
id3c_data = pd.read_csv(args.id3c_data, dtype = 'string')
id3c_data['qrcode'] = normalize_barcode(id3c_data['qrcode'])

The only thing that feels unnatural is that we're putting a generating PDF concern into the transform script.
Without this check we have to make sure the SQL logic sets a value for `pre_analytical_specimen_collection` in all cases.

Is it worth making transform handle generate PDF's concern here?

bin/return-of-results/transform
jccraft-uw: But it would really suck if the deployment actually turned the job back on after we manually stopped it. I've added that as a commit in here.

bin/return-of-results/transform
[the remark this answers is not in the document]
kschwabefry: Thanks for the details! I can see the behavior you describe now when I use append with/without setting `ignore_index=True` .

bin/return-of-results/transform
jccraft-uw: @kschwabefry thanks for talking through this. I've added a comment above `require_contacted_lamba` to explain why I'm comparing against pd.NA before comparing the string values.

bin/return-of-results/generate-pdfs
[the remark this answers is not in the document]
jccraft-uw: I've added this simple data quality check in as a new commit.

bin/return-of-results/transform
```diff
@@ -227,6 +227,23 @@ def edit_status_code(results: pd.DataFrame) -> pd.DataFrame:
...
+    """
+
+    # Check the `pre_analytical_specimen_collection` column
+    # pd.NA is included by ~ isin
+    no_pre_analytical_specimen_collection = joined_data[~joined_data['pre_analytical_specimen_collection'].isin \
+        (['IRB', 'clinical'])]
+
+    if not no_pre_analytical_specimen_collection.empty:
+        print(f"{len(no_pre_analytical_specimen_collection)} records have a bad pre_analytical_specimen_collection value. " \
+        f"They are qrcodes: {list(no_pre_analytical_specimen_collection['qrcode'])}", file = sys.stderr)
```
kschwabefry, line 244: Thanks for adding this check. nonblocker, just my two cents: If you want to make the logging more verbose, it could be also helpful to include no_pre_analytical_specimen_collection['pre_analytical_specimen_collection']

jccraft-uw: Totally! Anything that can keep us from having to dig through logs are run queries to see what the issue is. I'm now logging the qrcode and pre_analytical_specimen_collection.

kschwabefry (approved):

jccraft-uw:
Deployed.

PR #129, opened 2021-02-08
crontabs: Move promjob after fatigue

tsibley:
Fatigue masks failed exit statuses as part of its collab with chronic,
which means that promjob reports jobs as exiting with success when
fatigue was suppressing a repeated error.  This is unhelpful behaviour
for metrics and monitoring!, so move promjob after fatigue so it can see
real exit statuses.  If chronic and fatigue are paired, they are kept
together.  Uses of chronic sans fatigue were left alone, as chronic does
not modify exit statuses without -e, which we don't use.

I'd originally made promjob the very outer layer in the exec chain so
that it reported on the entire job, not just some inner piece of it, and
I think still a good practice to place it as left-most as reasonable.
In any case, we still expect to ditch fatigue once our alerting from
metrics and logs is better.


kschwabefry (approved):

jccraft-uw (approved):

tsibley:
Deployed.

PR #130, opened 2021-02-09
Remove bad assumption and combine append commands

jccraft-uw:
My original statement about appending DataFrames that have different
columns was incorrect. The values in the disconcordant columns will be
`NA` for a string column (and `NAN` for other types). There's no need
to replace data after the append. In fact, the `replace` line was
causing this warning:
`FutureWarning: elementwise comparison failed; returning scalar instead,
but in the future will perform elementwise comparison`

Also, the .append() method can take a list of DataFrames to append. It's
more efficient to append the two in the same command.

### DIFF
--- a/bin/return-of-results/transform
+++ b/bin/return-of-results/transform
@@ -267,12 +267,7 @@ if __name__ == '__main__':
 
     # Combine the DataFrames by appending to the first.
     # Use `ignore_index = True` to prevent duplicate index values.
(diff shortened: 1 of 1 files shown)


jccraft-uw:
[Fix RoR FutureWarning warning introduced last week](https://trello.com/c/IB32cl4X/1006-fix-ror-futurewarning-warning-introduced-last-week)

kschwabefry (approved):

jccraft-uw:
Deployed.

PR #132, opened 2021-02-16
wa-doh-linelists: Fix HCT zipcode

kairstenfay:
I was using the id3c-customizations `zipcode_map()` function to convert
REDCap raw zipcode labels to their real zipcode equivalent. However, HCT
zipcodes (at least in the EH&S project) are already the real zipcode. I
overlooked the fact that the `zipcode_map()` function returns None if
the input data don't exist in the map. This commit preserves the
original zipcode if no appropriate mapping was found (rather than
overwriting it with None).

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -54,7 +54,7 @@ def relabel_values(redcap_report: pd.DataFrame, config: Dict[str, Any]) -> pd.Da
     Re-assigns raw values a meaningful label in a given *redcap_report* using
     logic in id3c-customizations and the provided *config*.
     """
-    if 'home_zipcode_2' in redcap_report:
+    if config.get('relabel_zipcodes'):
         redcap_report['home_zipcode_2'] = redcap_report['home_zipcode_2'] \
             .apply(zipcode_map)
 
--- a/etc/wa-doh-linelists.yaml
+++ b/etc/wa-doh-linelists.yaml
@@ -151,6 +151,7 @@ column_rename_map: {
 static_columns: {
   "study_arm": "SCAN",
 }
+relabel_zipcodes: True
 
 ---
 name: SCAN IRB Spanish
@@ -170,6 +171,7 @@ static_columns: {
   "study_arm": "SCAN",
   "primary_language": "Spanish",
 }
+relabel_zipcodes: True
 
 ---
 name: SCAN IRB Vietnamese
@@ -182,6 +184,7 @@ static_columns: {
   "study_arm": "SCAN",
   "primary_language": "Vietnamese",
 }
+relabel_zipcodes: True
 
 ---
 name: SCAN IRB Russian
@@ -201,6 +204,7 @@ static_columns: {
   "study_arm": "SCAN",
   "primary_language": "Russian",
 }
+relabel_zipcodes: True
 
 ---
 name: SCAN IRB Traditional Chinese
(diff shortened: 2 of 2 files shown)


bin/wa-doh-linelists/transform
```diff
@@ -55,8 +55,10 @@ def relabel_values(redcap_report: pd.DataFrame, config: Dict[str, Any]) -> pd.Da
     logic in id3c-customizations and the provided *config*.
     """
     if 'home_zipcode_2' in redcap_report:
+        # Preserve the original zipcode even if the mapping does not exist in
+        # our zipcode_map() function (primarily intended for SCAN)
         redcap_report['home_zipcode_2'] = redcap_report['home_zipcode_2'] \
-            .apply(zipcode_map)
+            .apply(lambda zipcode: zipcode_map(zipcode) or zipcode)
```
jccraft-uw, line 61: It's not safe to mix mapped zipcodes and unmapped zipcodes. Imagine that we have this in the map:
98101: 98122. 
In that example, the key in the map might actually match a real zip code value. 
This line: .apply(lambda zipcode: zipcode_map(zipcode) or zipcode)
could erroneously use a mapped value with an input zipcode value of "98101" when "98101" was the zipcode itself.

kairstenfay: Fair point. Would it be better to only apply `zipcode_map()` to projects whose `name` starts with "SCAN"? I also considered this option at one point, but I don't love partial string matching if I can avoid it.

jccraft-uw: Whether the zipcode is mapped or not for the project could be added to wa-doh-linelists.yaml.

kairstenfay: Great idea @jccraft-uw . I just re-pushed using this suggestion.

kairstenfay:
[Linelists: Fix HCT zipcodes](https://trello.com/c/6KiR3sg9/1016-linelists-fix-hct-zipcodes)

kschwabefry (approved):

bin/wa-doh-linelists/transform
```diff
@@ -54,7 +54,7 @@ def relabel_values(redcap_report: pd.DataFrame, config: Dict[str, Any]) -> pd.Da
     Re-assigns raw values a meaningful label in a given *redcap_report* using
     logic in id3c-customizations and the provided *config*.
     """
-    if 'home_zipcode_2' in redcap_report:
+    if config.get('relabel_zipcodes'):
         redcap_report['home_zipcode_2'] = redcap_report['home_zipcode_2'] \
             .apply(zipcode_map)
```
jccraft-uw, line 59: `zipcode_map` is SCAN specific. It's coming from seattleflu.id3c.cli.command.etl.redcap_det_scan.
The map itself could be different across different projects. (There's no reason that value 123 should point to the same zip code across REDCap projects.)
Would there be a case where a non-SCAN project's values get passed in to the SCAN map?

kairstenfay: SCAN is the only project where we're mapping zipcodes, but if we did introduce a new mapping function, then you're right -- this code would need to be updated.

I don't see how non-SCAN zipcodes could be passed into the SCAN map unless the accompanying config file is update to set `relabel_zipcodes: True`.

jccraft-uw: For some reason, the Apple study and Snohomish schools study are mapping zip codes. Childcare and Adult Family Homes aren't.
We'll need to keep that in mind before we add projects to this application.

PR #134, opened 2021-02-17
Add Apple Respiratory Study to RoR

jccraft-uw:


jccraft-uw:
[Include Apple study in RoR pipeline](https://trello.com/c/bye2mILp/973-include-apple-study-in-ror-pipeline)

etc/sfs-longitudinal-redcap-projects.yaml
```diff
@@ -134,3 +134,24 @@ encounter_event_names:
...
+  - core_birthdate
+encounter_event_names:
+  - baseline_test_arm_1
+  - illness_episode_arm_1
+  - serial_event_1_arm_1
+  - serial_event_2_arm_1
+  - serial_event_3_arm_1
+  - serial_event_4_arm_1
+encounter_fields:
+  - collection_barcode
```
kschwabefry, line 156: If this stays as the variable name (see comment on the switchboard PR), I think we need to update `parse_sfs_longitudinal_redcap` to incorporate it. It looks like its currently working with

```
    barcode_fields = {
        "collect_barcode_kiosk",
        "return_utm_barcode"}
```

jccraft-uw: @kschwabefry Wow! Thanks for pointing out that I failed to add the barcode in there.
instead of adding it there (and then for the next time adding yet another barcode field in), I'm going to make the longitudinal config driven parsing work how the classic config driven parsing work. The configuration for the project will say in which order which barcode fields should be examined to end up with a "prioritized" barcode.
I'm closing this PR in favor of the new one.

PR #136, opened 2021-02-18
Add longitudinal mapping

jccraft-uw:
The bulk of this PR is adding field mapping and barcode prioritization logic to the SFS longitudinal configuration driven parser to match what was put in for the classic mode. This accommodates REDCap projects that use fields that aren't named like other projects and explicitly configures whether the project requires contact the participant before releasing positive results.
A commit fixes two small bugs in the classic mode.
A commit adds longitudinal configuration for the Apple Respiratory Study.

For testing, I ran `generate-results-csv` using this version of the code and right after using the master version of the code. A diff revealed no differences.

### DIFF
--- a/etc/sfs-longitudinal-redcap-projects.yaml
+++ b/etc/sfs-longitudinal-redcap-projects.yaml
@@ -3,15 +3,16 @@ redcap:
   url: https://redcap.iths.org
   project: 23854
 source: uw_reopening
+contact_before_releasing_result: false
+contacted_field: null
+participant_first_name_field: core_participant_first_name
+participant_last_name_field: core_participant_last_name
+participant_birthdate_field: core_birthdate
 enrollment_event_names:
   - enrollment_arm_1
-enrollment_fields:
-  - core_participant_first_name
-  - core_participant_last_name
-  - core_birthdate
 encounter_event_names:
   - encounter_arm_1
-encounter_fields:
+prioritized_barcode_fields:
   - collect_barcode_kiosk
   - return_utm_barcode
 ---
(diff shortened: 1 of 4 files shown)


jccraft-uw:
[Include Apple study in RoR pipeline](https://trello.com/c/bye2mILp/973-include-apple-study-in-ror-pipeline)

etc/sfs-longitudinal-redcap-projects.yaml
```diff
@@ -3,29 +3,32 @@ redcap:
...
 enrollment_event_names:
   - enrollment_arm_1
-enrollment_fields:
-  - core_participant_first_name
-  - core_participant_last_name
-  - core_birthdate
 encounter_event_names:
   - encounter_arm_1
-encounter_fields:
+additional_fields:
```
kschwabefry, line 15: We could leave out this key if its not relevant to the project's config.

kschwabefry (approved):
Nice to see `export-redcap-sfs-longitudinal `and `classic` alignment. In the future (not required for part of this), I wonder if there's some shared code we could pull out between those too, for example, that `SFSProject` class in both of those is really similar now.

etc/sfs-longitudinal-redcap-projects.yaml
jccraft-uw: That's a great point. I've removed it for all the longitudinal configs. No need to have an empty setting there.

jccraft-uw:
> Nice to see `export-redcap-sfs-longitudinal `and `classic` alignment. In the future (not required for part of this), I wonder if there's some shared code we could pull out between those too, for example, that `SFSProject` class in both of those is really similar now.

There is quite a bit of duplicated functionality between classic and longitudinal. That could be refactored in the future for sure.

jccraft-uw:
Deployed.

PR #138, opened 2021-02-26
Handle empty REDCap recordsets in config driven RoR

jccraft-uw:
The code needs to handle the case where the config driven
classic or longitudinal REDCap project fetch returns no records.
When processing the output of the fetch, check if the DataFrame
created from loading the data has rows, and if not, return None
instead of trying to transform that DataFrame.

### DIFF
--- a/bin/return-of-results/transform
+++ b/bin/return-of-results/transform
@@ -7,6 +7,7 @@ portal.
 import sys
 import argparse
 import pandas as pd
+from typing import Optional
 
 
 def parse_scan_redcap(redcap_file) -> pd.DataFrame:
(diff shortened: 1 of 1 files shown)


tsibley (approved):
LGTM by inspection. I assume you've run and tested it. :-) Thanks for digging into this.

jccraft-uw:
> LGTM by inspection. I assume you've run and tested it. :-) Thanks for digging into this.
Thanks for taking a look, @tsibley !
I did run locally and made sure my final output included results from SCAN and some longitudinal projects. We don't have data for the classic projects yet.

jccraft-uw:
Deployed.

PR #142, opened 2021-03-04
generate-results-csv: Kill all failed processes

kairstenfay:
Handle all process substitutions similarly, instructing the child
process to kill the parent in case it fails. This introduces a race
condition where the parent process will no longer be around for the non-
winning processes.

------------------------------

Addresses post-merge feedback from https://github.com/seattleflu/backoffice/pull/111.


kairstenfay:
[(Kairsten) respond to Tom's backoffice PR feedback](https://trello.com/c/btJAAYX3/962-kairsten-respond-to-toms-backoffice-pr-feedback)

tsibley (approved):

PR #144, opened 2021-03-12
Process REDCap data in chunks for the EH&S Transfer

jccraft-uw:
Previously, the entire output file from the REDCap dump was loaded
at once. This was causing out of memory errors, as the working memory
grew past 9 GB. In this commit, the REDCap file is read in chunks and
the content of each chunk is processed completely as one batch. Peak
memory consumption as configured is around 3.5 GB.

The original `transform` script took about 51 seconds to process today's
data. This new version took slighly longer at 63 seconds.

I've tested by comparing the output generated by the master version of `transform` against this new version. There are no differences in the output.

### DIFF
--- a/bin/uw-ehs-report/transform
+++ b/bin/uw-ehs-report/transform
@@ -4,60 +4,17 @@ Join the REDCap NDJSON data with an ID3C CSV.
 Prepares the joined data for import into the
 EH&S Transfer REDCap project.
 """
-import sys
+
 import argparse
+import io
+import json
 import pandas as pd
+import sys
+from typing import List
+
 from id3c.cli.io.pandas import dump_ndjson
 
-
-def parse_redcap(redcap_file) -> pd.DataFrame:
-    """
-    Reads in data from a given *redcap_file*. Returns a pandas.DataFrame
-    with columns renamed for the EH&S import.
-    """
-    redcap_data = (
-        pd.read_json(redcap_file, lines = True, dtype = False, convert_dates = False)
-        .astype("string")
-        .pipe(trim_whitespace)
-        .replace({"": pd.NA})
-        .astype("string")
-        .rename(columns={"dorm_room": "uw_dorm_room_number",
-        "sea_employee_type_other": "employee_category_other",
-        "dorm": "uw_dorm_name",
-        "uw_apt_names": "uw_apartment_name",
-        "core_participant_first_name": "participant_first_name",
-        "core_participant_last_name": "participant_last_name",
-        "core_birthdate": "birthdate",
-        "core_home_street": "home_street",
-        "core_apartment_number": "apartment_number",
-        "core_home_city": "home_city",
-        "core_home_state": "home_state",
-        "core_zipcode": "zipcode",
-        "pronouns_other": "preferred_pronouns_other",
-        "core_sex_other": "sex_other"
-        }))
-
-
-    # Normalize all barcode fields upfront.
-    barcode_fields = {
-        "collect_barcode_kiosk",
-        "return_utm_barcode"}
-
-    for barcode_field in barcode_fields:
-        if barcode_field in redcap_data:
-            redcap_data[barcode_field] = normalize_barcode(redcap_data[barcode_field])
-        else:
-            redcap_data[barcode_field] = pd.Series(dtype = "string")
-
-    kiosk_barcodes = redcap_data["collect_barcode_kiosk"]
-    mail_barcodes = redcap_data["return_utm_barcode"]
-
-    barcodes = kiosk_barcodes.combine_first(mail_barcodes)
-    barcodes = drop_duplicate_values(barcodes)
-    redcap_data['barcode'] = barcodes
-
-    enrollments = redcap_data.query('redcap_event_name == "enrollment_arm_1"')[[ \
-        "record_id",
+ENROLL_FIELDS =  [
         "netid",
         "affiliation_other",
         "uw_school",
@@ -79,82 +36,18 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
         "sex_other",
         "uw_greek_house",
         "uw_dorm_name",
-        "uw_apartment_name"
-        ]]
+        "uw_apartment_name"]
 
-    # This invariant protects our filename assumptions.
-    assert all(enrollments['birthdate'].str.match(r"^\d{4}-\d{2}-\d{2}$").dropna())
-
-    # Drop enrollment records with duplicated NetIDs.
-    netids = enrollments['netid']
-    netids = drop_duplicate_values(netids)
-    enrollments['netid'] = netids
-    enrollments.dropna(subset={'netid'}, inplace=True)
-
-    encounters = redcap_data.query('redcap_event_name == "encounter_arm_1"')[[ \
-        'record_id',
-        'barcode',
-        'illness_kiosk',
-        'illness_swabsend'
-        ]]
-
-    # Set symptomatic_when_tested on encounters.
-    # Use '1' and '0' values to be consistent with the other boolean values we send to REDCap.
-    encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
-    encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
-
-    joined_data = enrollments.merge(encounters, how='inner', on='record_id')
-
-    return joined_data
+# The number of lines to read from the REDCap file in one batch
+BATCH_SIZE = 10000
 
 
-def normalize_barcode(barcode):
-    if barcode.empty:
-        return pd.NA
-    return barcode.str.upper().str.strip()
-
-
-def trim_whitespace(df: pd.DataFrame) -> pd.DataFrame:
+def prepare_id3c_data(id3c_data_path: str) -> pd.DataFrame:
     """
-    Trim leading and trailing whitespace from strings in *df*.
+    Reads the file containing results data from ID3C.
+    Prepares the data for joining to REDCap data later.
     """
-    str_columns = df.select_dtypes("string").columns
-    df[str_columns] = df[str_columns].apply(lambda column: column.str.strip())
-    return df
-
-def drop_duplicate_values(input_series: pd.Series) -> pd.Series:
-    """
-    Find and drop duplicate values from the provided series and returns
-    the deduplicated pandas Series.
-    Avoids printing PII only if the given *input_series.name* is `netid`.
-    """
-    deduplicated_series = input_series
-
-    dups = input_series.loc[input_series.duplicated(keep = False)].dropna()
-    if not dups.empty:
-        deduplicated_series = input_series.drop(dups.index, inplace = False)
-
-        dups_count = len(dups)
-        dups_unique = list(dups.unique())
-        print(f"Dropped {dups_count:,} records with duplicated {input_series.name}: "
-            f"{dups_unique if input_series.name != 'netid' else '<masked>'}", file = sys.stderr)
-
-    return deduplicated_series
-
-if __name__ == '__main__':
-    parser = argparse.ArgumentParser(
-        description="Join a REDCAP export NDJSON file with a SCAN return of results ID3C export CSV file."
-    )
-    parser.add_argument("redcap_data", help="NDJSON export of SCAN records from REDCap")
-    parser.add_argument("id3c_data", help="CSV export of SCAN return of results from ID3C")
-    parser.add_argument("output", help="A destination for the output NDJSON", nargs="?",
-        default=sys.stdout, type=argparse.FileType("w"))
-
-    args = parser.parse_args()
-
-    redcap_data = parse_redcap(args.redcap_data)
-
-    id3c_data = pd.read_csv(args.id3c_data, dtype = 'string')
+    id3c_data = pd.read_csv(id3c_data_path, dtype = 'string')
     id3c_data['barcode'] = normalize_barcode(id3c_data['barcode'])
 
     # Drop id3c records with duplicated barcodes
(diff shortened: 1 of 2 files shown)


jccraft-uw:
[EH&S Transfer job is running out of memory. Figure out a redesign.](https://trello.com/c/EYx3ku7E/1049-ehs-transfer-job-is-running-out-of-memory-figure-out-a-redesign)

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+                            carryover_line = line
+                            break
+                    else:
+                        has_content = False
+                        break
+
+                content = io.StringIO(''.join(content))
+
+                redcap_data = (
+                    pd.read_json(content, lines = True, dtype = False, convert_dates = False)
```
tsibley, line 165: Why decode the JSON lines in the code above this part, but then decode them again from an accumulated string buffer instead of using the decoded records directly?

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+                        current_record_id = json.loads(line)['record_id']
+
+                        if current_record_id == last_record_id:
+                            content.append(line)
+                        else:
+                            carryover_line = line
+                            break
+                    else:
+                        has_content = False
+                        break
```
tsibley, line 160: This could be quite a bit simpler by iterating over the file handle by lines instead of reading chunks. Chunks make a lot of sense for formats which aren't line-based, but the whole benefit of NDJSON is that it's line-based. :-) Doing so would avoid the need for tracking things like `has_content`, `carryover_line`, and even `content`; instead, it'd process a batch of input and dump output every N records/rows.

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+            # and reset carryover_line to None.
+            if carryover_line:
+                content.insert(0, carryover_line)
+                carryover_line = None
+
+            content.extend(f.readlines(CHUNK_BYTES))
+            has_content = len(content) > 0
+
+            if has_content:
+                last_record_id = json.loads(content[len(content) -1])['record_id']
```
tsibley, line 144: FYI, you can index into arrays from the end by using negative indexes:

```python
content[-1] # last element
content[-2] # second to last element
```

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
 
-    # Set symptomatic_when_tested on encounters.
-    # Use '1' and '0' values to be consistent with the other boolean values we send to REDCap.
-    encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
-    encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
+    return duplicates
+
+
+def process_redcap_data_and_dump_output(redcap_file: str, id3c_data: pd.DataFrame,
+        output: str) -> pd.DataFrame:
```
tsibley, line 108: `output` is not a string, but an `argparse.FileType`.

bin/uw-ehs-report/transform
```diff
@@ -152,57 +324,5 @@ if __name__ == '__main__':
...
-    ]
-
-    for field in boolean_fields:
-        id3c_data[field].replace({'t': 1, 'f': 0}, inplace=True)
-
-    # Only keep records we can match from both REDCap and ID3C.  An incorrect
-    # barcode in REDCap or the lack of a record in ID3C may cause records to
-    # drop out in the join, but this is preferable to attaching results to an
-    # incorrect barcode or not having a known status for a barcode.
-    joined_data = redcap_data.merge(id3c_data, how='inner', on='barcode')
```
tsibley, line 185: This comment was dropped when the join was moved, but I think it's useful context which should be kept in the new location.

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+                encounters.loc[(encounters['illness_kiosk'] == 'yes') | (encounters['illness_swabsend'] == 'yes'), 'symptomatic_when_tested'] = '1'
+                encounters.loc[(encounters['illness_kiosk'] == 'no') | (encounters['illness_swabsend'] == 'no'), 'symptomatic_when_tested'] = '0'
+
+                encounters = enrollments.merge(encounters, how='inner', on='record_id')
+                encounters = encounters.merge(id3c_data, how='inner', on='barcode')
+
+                dump_joined_data(encounters, output)
+
+
+def dump_joined_data(joined_data: pd.DataFrame, output: str) -> None:
```
tsibley, line 254: Ditto re: the type of `output`.

tsibley (commented):
Thanks for tackling this. I noted a few places where I think the logic could be quite a bit simpler by using a different approach.

As a rule, I like checking data diffs to check big changes like this one. Have you diffed the output from `master` vs. this branch when given the same input?

jccraft-uw:
> Previously, the entire output file from the REDCap dump was loaded
> at once. This was causing out of memory errors, as the working memory
> grew past 9 GB. In this commit, the REDCap file is read in chunks and
> the content of each chunk is processed completely as one batch. Peak
> memory consumption as configured is around 3.5 GB.
> 
> The original `transform` script took about 51 seconds to process today's
> data. This new version took slighly longer at 63 seconds.
> 
> I've tested by comparing the output generated by the master version of `transform` against this new version. There are no differences in the output.

I noted this in the PR description and just added it to commit message:
'I've tested by comparing the output generated by the master version of transform against this new version. There are no differences in the output.'
I love having a known good product (the master branch) to compare against.

jccraft-uw:
> Thanks for tackling this. I noted a few places where I think the logic could be quite a bit simpler by using a different approach.
> 
> As a rule, I like checking data diffs to check big changes like this one. Have you diffed the output from `master` vs. this branch when given the same input?

I noted this in the PR description and just added it to commit message:
'I've tested by comparing the output generated by the master version of transform against this new version. There are no differences in the output.'
I love having a known good product (the master branch) to compare against.

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+    more than one record_id.
+    """
+    found_netids = {}
+    duplicates = []
+
+    with open(redcap_file, 'r') as f:
+       while True:
+           line = f.readline()
+
+           if line:
```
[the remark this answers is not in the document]
jccraft-uw: WHOO indeed! The Python way is much nicer. I've updated that.

bin/uw-ehs-report/transform
jccraft-uw: I'm using chunks as a fast way to grab ~CHUNK_BYTES worth of data from the file. 
`f.readlines(CHUNK_BYTES)` (passing CHUNK_BYTES to the `sizehint` parameter) will read ~CHUNK_BYTES bytes worth of data from the file and will include a whole line.

If the `readlines` does return content, I look at the `record_id` in the last line read and then read lines (in the `while` loop) until I have read a line with a different record_id. I need to do this because in the NDJSON file we have a row per record instance (not a row per record_id).  I need to make sure that each batch of processing includes all the instances for the records because I'm joining the enrollment instance to encounter instances. For example, if the chunk left an encounter instance without any other rows for that record_id, this would be processed in the next batch, and it would drop out because it wouldn't have its enrollment instance to join to.

bin/uw-ehs-report/transform
jccraft-uw: I only decode the JSON of the lines that I need to iterate through after I have a chunk of data (after `f.readlines(CHUNK_BYTES)`). I decode each of the additional lines so that I can see the line's record_id value to know if I need to keep going. These are a small fraction of the lines in `content`.
`pd.read_json(content ...)` does decode all the lines in `content`.

bin/uw-ehs-report/transform
jccraft-uw: Updated!

bin/uw-ehs-report/transform
jccraft-uw: Agreed. Thanks. I added it back.

bin/uw-ehs-report/transform
```diff
@@ -79,33 +36,246 @@ def parse_redcap(redcap_file) -> pd.DataFrame:
...
+    has_content = True
+
+    # We want to include all lines for a particular `record_id` in the same batch so that
+    # we always have the record_id's enrollment instance when we process its encounter
+    # instances. After we read a chunk, we will read lines until we find a line with a record_id
+    # that is different than the record_id of the record at the end of the chunk. This line is
+    # a "carry over" line that gets added at the start of the next batch.
+    carryover_line = None
+
+    with open(redcap_file, "r") as f:
```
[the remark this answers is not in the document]
jccraft-uw: That seems kind of important, doesn't it?!?
I've updated generate-report to write the REDCap and ID3C results data to temp files. This means that we're not pulling those 2 data sets in parallel. When I ran generate-report locally just now the time goes from 8 minutes (using what's in Master) to 11 minutes (using this branch, using the temp files and chunking through REDCap.)

bin/uw-ehs-report/transform
jccraft-uw: Updated.

bin/uw-ehs-report/transform
jccraft-uw: Updated.

jccraft-uw:
@tsibley , thanks again for reviewing this. I have addressed all your comments and questions.

bin/uw-ehs-report/transform
tsibley: Ah, I think you're maybe missing my point, which means I didn't express it well. I fully understand what the current code is doing and why it needs to track what it does (but thank you for explaining it in your comment; it confirms our understanding matches).

My point was that the code has to do all that because of the decision it makes to read data in arbitrary chunks of lines instead of chunks of record ids (read line-by-line). It essentially inserts an additional and unnecessary buffer/queue between the file and the data manipulation code (everything from `pd.read_json(…)` down).

Maybe it'll be more helpful to sketch out what I mean:

```python
batch_size = 5000
rows = []

for row in (json.loads(line) for line in file):
    if len(rows) >= batch_size and row["record_id"] != rows[-1]["record_id"]:
        # We've made our batch size and are at a record boundary.  Process the
        # current batch now.
        process(rows)
        rows = []

    rows.append(row)
else:
    # Pick up the last batch
    if rows:
        process(rows)

def process(rows):
    redcap_data = (
        pd.read_records(rows)
        ...)

    dump_joined_data(...)
```

jccraft-uw: Thanks for all that.
The fundamental difference is that with the current implementation we control memory utilization. We grab about 200 MB of data from the file. (It will be a little larger if we need to keep reading lines to get at a record boundary.)

With the proposed approach we have to decide on a batch_size and we might need to decrease that over time as the number of record instances per record grows. We would have to keep an eye on memory utilization.

tsibley: Hmm, I think both control memory utilization very similarly without needing continual adjustment. Both `CHUNK_BYTES` and `batch_size` correlate with the number of lines, or record instances, to process in one go. As the number of record instances per record id grows, both will process fewer record ids in each chunk/batch while the total size of the chunk/batch remains similar. If each record instance/row/line started being larger (e.g. the number of fields per record was growing or the values of the existing fields was systematically growing), then a difference between the two approaches would start to arise. This seems to me to be not much of a concern as I don't think individual record size is growing quickly (do we have data that could tell us?). To me, it is is outweighed by not needing to track a bunch of fiddly state.

jccraft-uw: That's right. I mistakenly thought of batch_size as number of records and not number of rows.

I've already taken care of tracking the state, so that's not so fiddly. If you'd prefer I go with your approach I will.

tsibley: I prefer the approach I suggested, as I think it is clearer and simpler. I know this PR contains working code, but I think the approach it takes imposes more effort to understand what the code is doing and requires a reader to keep track of more things.

My guiding principle here is that the cost of the "fiddly" bits is not just in _writing_ them, but also in _reading_ them. Code is read many more times than it is written, and maintainability is very important when folks inevitably have to come back to the code later. This matters to me just as much (if not more) than merely having something that works at the moment.

tsibley: I'll add that pragmatism is also a good guiding principle, and intersects with what I said above. If the memory issues with this job were urgent enough, I'd say merge the working code in this PR and simplify/improve it later. It seems that they're not very urgent, though, and so my inclination is to take the time to make the improvements now.

jccraft-uw: @tsibley , the suggested code works really well and, agreed, is easier to read. I took your suggested code and incorporated it with these changes:
1. I set the BATCH_SIZE to 10000. I tried various values and found this to be the sweet spot (running on my laptop) for run time and memory consumption.
2. I'm not loading the JSON for every line. I load the JSON only after our len(rows) reaches BATCH_SIZE. We don't need the other lines as JSON, so I wanted to save this bit of processing.

I also fixed the return type for process_redcap_data_and_dump_output. It doesn't return anything, so None is the correct type.

See what you think with the latest commit. I have tested, comparing the transformed output with this code against that generated by the Master version, and they match. I did that twice, with about an hour and a half between.

tsibley: > I'm not loading the JSON for every line. I load the JSON only after our len(rows) reaches BATCH_SIZE. We don't need the other lines as JSON, so I wanted to save this bit of processing.

Ah, I'm not sure there's any processing saved here. The `pandas.read_json()` call will deserialize every line once, but also some lines are deserialized an additional 2 times prior to that when testing for the record id boundary. I don't expect this makes any meaningful difference, but in the example I wrote, each line is only ever deserialized once (by the generator that's looped over).

One other observation: I think the memory consumption of this code will be 2x higher per batch than it strictly needs to be because the code maintains both an array of lines (`rows`) as well as a string concatenation of those lines (`batch`) at the same time. If batches of lines are desired, then it seems it'd be more memory efficient to use two variables: an appended-to string of lines and a counter of lines seen so far. (I didn't use a separate counter in my example because there was not the same duplication of data, but it also probably has higher base memory usage since it stores dictionaries rather than strings.)

tsibley (approved):
LGTM. Thanks for reworking and testing it! I left a couple observations in a comment, but I don't think they have to block merge unless you want to address them.

bin/uw-ehs-report/transform
jccraft-uw: It took a hammer hitting me over the head before I realized you were using a different method to load the data frame. 
I've switched to using `pd.DataFrame.from_records(rows)` to load the rows (as list[dict]), so now I no longer have two copies of the data. That's very nice.

Given that we can load a list of dictionaries into the data frame, I switched to how you were using the generator to convert to JSON only once.

I did another comparison against the output produced by the code in the Master branch as a sanity check. They match.

jccraft-uw:
Deployed.

PR #145, opened 2021-03-16
Wrap refresh materialized view jobs into one job

jccraft-uw:
The `__uw_encounters` and `scan_encounters_v1` materialized views depend
 on the `fhir_questionnaire_responses_v1` materialized view.

 This commit adds a new script that refreshes the materialized shipping
 views in the correct order and updates the production crontab to
 run this script instead of refreshing the materialized views
 individually.

 Currently, it takes about 10 minutes to materialize the 3 views in
 production. Therefore, the refresh job is configured to run every 30
 minutes. The start times of the `UW Reopening testing offers` job have
 been pushed back to allow the materialized views refresh to take
 up to 25 minutes.

### DIFF
--- a/bin/refresh-materialized-shipping-views
+++ b/bin/refresh-materialized-shipping-views
@@ -1,0 +1,59 @@
+#!/bin/bash
+# usage: ./bin/refresh-materialized-shipping-views
+#
+# Refreshes materialized views in the `shipping` schema
+# in the order configured.
+#
+
+set -euo pipefail
+
+
+# The materialized views to refresh.
+# They will get refreshed in the order listed.
+shipping_views=(
+    "fhir_questionnaire_responses_v1"
+    " __uw_encounters"
+    "scan_encounters_v1"
+    )
+
+
+main() {
+    for arg; do
+        case "$arg" in
+            -h|--help)
+                print-help
+                exit 0;;
+        esac
+    done
+
+    refresh-materialized-views
+}
+
+
+refresh-materialized-views() {
+
+    for view in "${shipping_views[@]}"; do
+        id3c refresh-materialized-view 'shipping' ${view} --commit
+    done
+
+}
+
+
+print-help() {
+    # Print the help comments at the top of this file ($0)
+    local line
+    while read -r line; do
+        if [[ $line =~ ^#! ]]; then
+            continue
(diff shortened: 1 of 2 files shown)


jccraft-uw:
[Refresh materialized shipping views: consolidate into one script](https://trello.com/c/Y7Skgp6p/1053-refresh-materialized-shipping-views-consolidate-into-one-script)

bin/refresh-materialized-shipping-views
```diff
@@ -0,0 +1,59 @@
...
+    done
+
+    refresh-materialized-views
+}
+
+
+refresh-materialized-views() {
+
+    for view in "${shipping_views[@]}"; do
+        id3c refresh-materialized-view 'shipping' ${view} --commit
```
kschwabefry, line 36: A question on how this handles failure cases: Do we want the other refreshes to continue if one fails? How strict is the dependency between refreshing these materialized views?

kschwabefry (approved):
Left one question, makes sense overall.

bin/refresh-materialized-shipping-views
jccraft-uw: Good question. The way it's written now, one failure will cause the whole job to cease and fail.
If the fhir_questionnaire_responses_v1 (the first one) fails it makes sense to stop since the other two depend on fhir_questionnaire_responses_v1.
If the __uw_encounters refresh fails it doesn't have to prevent the scan_encounters_v1 refresh from running.
I will figure out how to configure that in bash tomorrow.

jccraft-uw: I want to make sure that we log (and warn on) __uw_encounters's failure if we continue with scan_encounters_v1.
I need to see how that works with how our alerting works .... can it log without failing the job, for example.

kschwabefry: Agreed, I'd prioritize keeping logging/warning over continuing the refresh of the scan job. Honestly, adding complexity to the dependency tree may not even be worth it in this case - perhaps we really do just want any job failing to stop the whole job.

jccraft-uw: Let's keep it as is--one failure stops the whole job--and see if that causes issues that would make us want to handle failures differently.

jccraft-uw:
Deployed.

PR #152, opened 2021-04-01
Update UW retrospective manifest config with aliquot sheet 11

kschwabefry:
Lab is rolling over to aliquot manifest sheet 11, so start ingesting.

### DIFF
--- a/etc/uw-retrospectives-manifest.yaml
+++ b/etc/uw-retrospectives-manifest.yaml
@@ -198,3 +198,13 @@ extra_columns:
   mrn: mrn
   accession_no: accession
   sample_origin: sample_origin
+---
+workbook: s3://fh-pi-bedford-t/seattleflu/bbi/2020_2021_sfs_aliquoting_11.xlsx
+sheet: aliquoting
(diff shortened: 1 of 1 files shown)


etc/uw-retrospectives-manifest.yaml
```diff
@@ -198,3 +198,13 @@ extra_columns:
...
   accession_no: accession
   sample_origin: sample_origin
+---
+workbook: s3://fh-pi-bedford-t/seattleflu/bbi/2020_2021_sfs_aliquoting_11.xlsx
+sheet: aliquoting
+sample_column: sample_id
+extra_columns:
+  barcode: sample_id
+  collection_date: collection_date
+  mrn: mrnuw-
```
tsibley, line 208: `mrnuw-` looks like a typo?

tsibley (changes requested):

etc/uw-retrospectives-manifest.yaml
kschwabefry: Indeed. Fixed!

PR #159, opened 2021-05-13
Automate linelist generation and upload to S3

kschwabefry:
Use cron to generate daily linelists and upload to S3. Study staff can
then download the file from S3 to distribute to partner agencies,
instead of having to generate it locally.

This lets us control the environment for linelist generation and
simplifies the process for study staff. Eventually, it'd be great to
automatically upload the linelists directly to the partner agencies.

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -244,6 +244,6 @@ if __name__ == "__main__":
     mega_linelist = mega_linelist[universal_config['column_names']]
 
     mega_linelist.sort_values(by=['record_id']) \
-        .to_csv(args.output_dir / f"linelist_{date.replace('/', '')}.csv",
+        .to_csv(f"{args.output_dir}/linelist_{date.replace('/', '')}.csv",
             index=False,
             line_terminator='\r\n')
--- a/crontabs/id3c-production
+++ b/crontabs/id3c-production
@@ -130,3 +130,7 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 # Make UW Reopening testing offers at :15 and :45
 # This job uses data created by redcap-det uw-reopening, etl fhir, and materialized shipping views.
 25,55  * * * * ubuntu fatigue --quiet promjob "id3c offer-uw-testing" envdir $ENVD/redcap pipenv run id3c offer-uw-testing --commit
+
+# Generate and upload the previous day's linelist data once a day at 6 A.M.
+0 6 * * * ubuntu promjob "wa-doh-linelists" pipenv run envdir $ENVD/hutch envdir $ENVD/redcap wa-doh-linelists/generate
+--date $(date --date=yesterday --iso-8601=date) --output-dir "s3://fh-pi-bedford-t/seattleflu/public-health-reporting"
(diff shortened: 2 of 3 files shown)


kschwabefry:
[Automate Linelists: Help Kathryn, Cooper, Melissa, Brian](https://trello.com/c/ZwsYK8oA/1108-automate-linelists-help-kathryn-cooper-melissa-brian)

bin/wa-doh-linelists/transform
```diff
@@ -244,6 +244,6 @@ if __name__ == "__main__":
     mega_linelist = mega_linelist[universal_config['column_names']]
 
     mega_linelist.sort_values(by=['record_id']) \
-        .to_csv(args.output_dir / f"linelist_{date.replace('/', '')}.csv",
+        .to_csv(f"{args.output_dir}/linelist_{date.replace('/', '')}.csv",
```
tsibley, line 247: Oh, I forgot Pandas could also write directly to S3 and thought we'd have to use fsspec directly ourselves. That's great that we don't!

crontabs/id3c-production
```diff
@@ -130,3 +130,7 @@ GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
 # Make UW Reopening testing offers at :15 and :45
 # This job uses data created by redcap-det uw-reopening, etl fhir, and materialized shipping views.
 25,55  * * * * ubuntu fatigue --quiet promjob "id3c offer-uw-testing" envdir $ENVD/redcap pipenv run id3c offer-uw-testing --commit
+
+# Generate and upload the previous day's linelist data once a day at 6 A.M.
+0 6 * * * ubuntu promjob "wa-doh-linelists" pipenv run envdir $ENVD/hutch envdir $ENVD/redcap wa-doh-linelists/generate
+--date $(date --date="yesterday" +%F) --output-dir "s3://fh-pi-bedford-t/seattleflu/public-health-reporting"
```
tsibley, line 136: `%` are one of the gotchas in crontabs.

> The ``sixth'' field (the rest of the line) specifies the command to  be run.   The  entire  command  portion  of the line, up to a newline or % character, will be executed by /bin/sh or by the shell specified in the SHELL  variable of the crontab file.  Percent-signs (%) in the command, unless escaped with backslash (\\), will be changed into newline characters,  and  all  data  after the first % will be sent to the command as standard input. There is no way to split a  single  command  line  onto multiple lines, like the shell's trailing "\\".

So we'll want to make that `date` command either:

    date --date=yesterday +\%F

or equivalently, without risk of `%` weirdness:

    date --date=yesterday --iso-8601=date

tsibley (changes requested):
I'm glad this change ended up being so small! Clearly we should have done this sooner. :-)

One minor change to the crontab syntax, otherwise this LGTM.

bin/wa-doh-linelists/transform
kschwabefry: Yeah!! When you mentioned fsspec in trello, I looked into this and was very happy to see pandas already supported it!

crontabs/id3c-production
kschwabefry: 👀 Thanks for catching that! Taking your second suggestion, just verified it works with the linux we're running on backoffice - I'll push a new commit shortly.

tsibley (approved):
Awesome, thanks!

kschwabefry:
Deployed.

PR #160, opened 2021-05-13
Exclude queries run by Metabase for sync/scan introspection from the slow-query killer

tsibley:
Otherwise some of these routine queries are not completing for very
field/value enumerations of very large and slow views.  We will
additionally reduce their cadence so they don't run as often.

Queries that are run on behalf of a user have an extended remark comment
at the beginning of the query indicating as much.¹  Positively select
for these to allow any non-user driven queries through.

¹ https://github.com/metabase/metabase/blob/b8c81d0277c509f37ade13a764eef5dc42a06fb6/src/metabase/query_processor/util.clj#L26-L37

### DIFF
--- a/bin/terminate-slow-metabase-queries
+++ b/bin/terminate-slow-metabase-queries
@@ -11,6 +11,7 @@ psql --quiet --no-align --tuples-only --set ON_ERROR_STOP= <<<"
             pid != pg_backend_pid()
             and usename in ('metabase', 'metabase-privileged')
             and state = 'active'
+            and query like e'-- Metabase:: userID:%'
         order by
             active_duration desc
     ),


kschwabefry (approved):

tsibley:
Deployed.

tsibley:
[Reverted](https://github.com/seattleflu/backoffice/commit/a4e5d216a7d9c8a274df2f5e4732b1f691df195c) this 5 min after deploy, as the `pg_catalog.pg_stat_activity.query` column isn't passed through our proxy view `public.pg_stat_activity_nonsuperuser`.

We could address this a few ways that come to mind:

1. Drop this change entirely, and do nothing more. Assume that the RDS instance size upgrade will make the internal Metabase queries run quick enough now to not matter. I think this is unlikely to be the case.
2. Run this process as superuser and switch to using `pg_stat_activity` directly. I don't like this at all because it leaves backoffice with access to a database superuser.
3. Add the `query` column to `pg_stat_activity_nonsuperuser`. I don't like this because we may not want to expose all queries to everyone.
4. Add a severely truncated (or masked, but how?) `query` column and match against that. Feels a bit hacky to me and maybe attempting to be too generic a solution, although leading first-line comments are a common convention.
5. Extract information in Metabase comment into a new structured column that's part of `pg_stat_activity_nonsuperuser`, e.g.:

```sql
hstore(
    array['user_id', 'query_type', 'query_hash'],
    regexp_match(
        query,
        e'^-- Metabase:: userID: (.+?) queryType: (.+?) queryHash: (.+?)\n'))
```

Of these, I think I prefer 5. Thoughts?

tsibley:
I implemented 5 as https://github.com/seattleflu/id3c-customizations/pull/214, but a small change will be needed in this repo as well, something like:

```
and metabase->'user_id' is not null
```

PR #164, opened 2021-06-04
Added check for missing values in essential columns of linelist

cooperqmarshall:
Log a warning if there are any NA values in columns: collection_date, sample_barcode, date_tested, and test_result of the results linelist to wa doh.

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -240,6 +240,13 @@ if __name__ == "__main__":
         log.info(f"Created linelist with {str(len(linelist))} rows.")
         mega_linelist = pd.concat([mega_linelist, linelist])
 
+    # Check if collection date, sample barcode, date tested, or test result
+    # columns are missing a value
+    if mega_linelist[['collection_date','sample_barcode','date_tested','test_result']].isnull().values.any():
+        log.warning('Value missing from required column in row: %s', \
+        mega_linelist[mega_linelist[['collection_date','sample_barcode','date_tested','test_result']] \
+        .isnull().any(axis=1)].index[0])
+
     # Reorder columns
     mega_linelist = mega_linelist[universal_config['column_names']]
 


bin/wa-doh-linelists/transform
```diff
@@ -240,6 +240,13 @@ if __name__ == "__main__":
         log.info(f"Created linelist with {str(len(linelist))} rows.")
         mega_linelist = pd.concat([mega_linelist, linelist])
 
+    # Check if collection date, sample barcode, date tested, or test result
+    # columns are missing a value
+    if mega_linelist[['collection_date','sample_barcode','date_tested','test_result']].isnull().values.any():
+        log.warning('Value missing from required column in row: %s', \
+        mega_linelist[mega_linelist[['collection_date','sample_barcode','date_tested','test_result']] \
+        .isnull().any(axis=1)].index[0])
```
tsibley, line 248: One immediate note I have is that we won't actually see these warnings from the automatic jobs because of how it runs (we configured it to only send output when the job fails). We can adjust how it runs or do something other than warn. Longer term, there's some in-progress work to bring better visibility into logs, which would let us alert on this particular line very easily without hard coding the alert into here.

cooperqmarshall: Do all errors get posted in #id3c-alerts? If so would removing "chronic" and increasing the LOG_LEVEL to warn work to get this alert through?

tsibley: Yes, but we'd also want to remove/conditionalize the `echo`'s in `generate`, because otherwise they will generate false "alerts". That's the reason we slapped `chronic` on there in the first place, but admittedly `chronic` is a large hammer.

seidelma: One suggestion--to improve readability and extensibility, you might want to make a separate list of required fields, in case that changes in the future. Maybe something like:

`required_fields = [
  'collection_date',
  'sample_barcode',
  'date_tested',
  'test_result'
]`

`log.warning('Value missing from required column in row: %s', \
  mega_linelist[mega_linelist[required_fields] \
    .isnull().any(axis=1)].index[0])
`

PR #165, opened 2021-06-08
Phskc linelist

cooperqmarshall:
Write another linelist csv for phskc that contains only those with king county zipcodes.


bin/wa-doh-linelists/transform
```diff
@@ -254,3 +254,22 @@ if __name__ == "__main__":
...
+                '98056','98059','98011','98028','98001','98002','98003','98023',
+                '98030','98031','98032','98042','98047','98055','98057','98058',
+                '98148','98166','98168','98188','98198','98010','98022','98038',
+                '98045','98051','98065','98027','98029','98052','98074','98075',
+                '98092','98070','98014','98077','98053','98024','98072','98019']
+    # Linelist generated for PHSKC (only king county zipcodes) in same directory
+    # as the mega_linelist
+    phskc_linelist = mega_linelist[mega_linelist['home_zipcode_2'].isin(king_zips)]
+    phskc_linelist.sort_values(by=['record_id']) \
+        .to_csv(f"{args.output_dir}/PHSKC_SCAN_linelist_{date.replace('/', '')}.csv",
```
tsibley, line 273: If this remains all studies and not just SCAN, then let's remove "SCAN" from the filename.

bin/wa-doh-linelists/transform
```diff
@@ -254,3 +254,22 @@ if __name__ == "__main__":
...
+    king_zips = ['98101','98102','98103','98104','98105','98106','98107','98108',
+                '98109','98112','98115','98116','98117','98118','98119','98121',
+                '98122','98125','98126','98133','98134','98136','98144','98146',
+                '98154','98155','98164','98177','98178','98195','98199','98004',
+                '98005','98006','98007','98008','98033','98034','98039','98040',
+                '98056','98059','98011','98028','98001','98002','98003','98023',
+                '98030','98031','98032','98042','98047','98055','98057','98058',
+                '98148','98166','98168','98188','98198','98010','98022','98038',
+                '98045','98051','98065','98027','98029','98052','98074','98075',
+                '98092','98070','98014','98077','98053','98024','98072','98019']
```
tsibley, line 268: As discussed during dev office hours, it'd be good to double-check (probably with Tina) that we can release results for anyone in King County to PHSKC directly.

tsibley (commented):
Overall approach is reasonable. :+1:

bin/wa-doh-linelists/transform
cooperqmarshall: Will remove!

cooperqmarshall:
```suggestion
        .to_csv(f"{args.output_dir}/PHSKC_linelist_{date.replace('/', '')}.csv",
```

cooperqmarshall:
PHSKC will coordinate with DOH to get the data they need. This PR is no longer needed

cooperqmarshall:
PHSKC has a DUA with SFS so they would prefer to get the complete results data from us and not DOH. Reopening because this process will be needed.

davereinhart:
One of the automated tests failed:
https://github.com/seattleflu/backoffice/pull/165/checks?check_run_id=3245091405#step:4:6

You'll want to use curly braces there to insert a variable into a string (e.g. `echo "printing ${variable}`"

davereinhart:
The commit history here can be simplified by using `git mv` to rename files and folders. That way they won't be treated as entirely new files, and the `diff` will be much smaller. That will help make the commit history cleaner and make it a lot easier to review (all of the renaming can probably be combined into a single commit). I'm happy to jump on a quick call to walk through that.

cooperqmarshall:
@davereinhart Oh that is a useful command! I will use that to consolidate these commits. I'll let you know if I need support.

cooperqmarshall:
The commits have been rebased and the git diff looks good to me now. @davereinhart

bin/health-dept-linelists/generate
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# usage: ./bin/health-dept-linelists/generate --date DATE --health-dept HEALTH DEPARTMENT
```
davereinhart, line 2: It looks like there are more than just these two optional arguments for this command. For anyone just looking to use the `generate` command, they should be made aware of all of the command line options here instead of having to look at the transform command to find them. So that would mean including all of transform command's optional and required arguments here, and enforcing the required ones.

bin/health-dept-linelists/transform
```diff
@@ -154,6 +155,11 @@ def fill_empty_columns(linelist: pd.DataFrame, config: Dict[str, Any]) -> pd.Dat
     ]
     return linelist.assign(**{ c: pd.NA for c in missing_columns })
 
+def filter_county(linelist: pd.DataFrame) -> pd.DataFrame:
+    with open('./etc/county-zipcode-list.yaml', 'r') as f:
+        zipcode_list = list(yaml.safe_load_all(f))
+        return linelist[linelist['home_zipcode_2'] \
+            .isin(zipcode_list[0][args.health_dept])]
```
davereinhart, line 162: You may want to add a check here to make sure `args.health_dept` is present in the `county-zipcode-list.yaml`, and if it's not then raise an exception with error message indicating that. Otherwise `zipcode_list[0][args.health_dept]` would just throw a more generic KeyError exception that would be need to be debugged.

davereinhart (changes requested):
Much cleaner @cooperqmarshall, thanks for making those changes! I just did a quick visual scan and had a couple of minor suggestions. I can do a more complete review with testing later today.

bin/health-dept-linelists/generate
cooperqmarshall: Is there specific notation between required and optional that I should use here e.g. --date <DATE> and [--date <DATE>]?

bin/health-dept-linelists/transform
cooperqmarshall: Could I do this check in `generate` with a `grep` command instead?

bin/health-dept-linelists/generate
davereinhart: Square brackets are used to indicate optional arguments. Here's a helpful reference for the proper syntax:
http://docopt.org/

bin/health-dept-linelists/transform
davereinhart: That would require hardcoding in `generate`, which would be another piece of code that needs to be updated any time a new health department is added. A better approach to keep it dynamic might be a try/except to catch a KeyError exception.
```
try:
    zipcodes = zipcode_list[0][args.health_dept]
except KeyError as key:
    print(f"Required key «{key}» missing from yaml file")
```
The other part that could be simplified here is the zipcode_list object which you're converting from an object to a list which will only have a single item. You should be able to use `yaml.safe_load(f)` instead of `yaml.safe_load_all(f)` so then you don't need to select the first item from a list on line 162.
```
health_dept_zipcodes = yaml.safe_load(f)
```
Then replace `zipcode_list[0]` in your code with `health_dept_zipcodes`.

bin/health-dept-linelists/generate
```diff
@@ -0,0 +1,100 @@
...
+# 
+# Mandatory agruments:
+# --date <DATE>
+# --health-dept <wa-doh|phskc>
+# 
+# Optional Arguments:
+# [-h|--help]
+# [--output-dir <output-directory>]
+# [--config-file <config.yaml>]
+# [--ulpoad-if-empty <output-directory>]
```
davereinhart, line 16: typo

bin/health-dept-linelists/transform
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
 # usage: transform [-h] --date <date> [--project-id <project ID>]
 #                         --id3c-data <id3c-data.csv>
+#                         --health-dept <wa-doh|phskc>
 #                         [--output-dir <output-directory>]
 #                         [--config-file <config.yaml>]
+#                         [--ulpoad-if-empty <output-directory>]
```
davereinhart, line 7: Same typo here, also consider changing argument name to match argparse metavar below.

cooperqmarshall: Ah good catch! I will make the corrections for both of the typo locations. I also will change the metavar to <always-upload-dest> as you suggested.

cooperqmarshall: Made the corrections to the typo and metavar in generate and transform

crontabs/health-dept-linelists
```diff
@@ -0,0 +1,28 @@
+SHELL=/bin/bash
+PATH=/opt/backoffice/bin:/usr/local/bin:/usr/bin:/bin
+
+# Point Pipenv at the production environment
+PIPENV_PIPFILE=/opt/backoffice/wa-doh-linelists/Pipfile
```
seidelma, line 5: I think we need to switch this over to `/opt/backoffice/health-dept-linelists/Pipfile`

cooperqmarshall: Yes forgot to update that reference. I will force push with the correction

seidelma:
Almost there--just set line 89 to `"${args[@]}"` to shut up the shellcheck test and then I think we're good to go.

PR #171, opened 2021-06-22
Publish weekly pathogen prevalence data every day

tsibley:
This data file will power a chart of pathogen prevalence on
seattleflu.org.

The s3://seattle-flu-study bucket existed already and was previously
used for the samples.geojson file powering the heat map on
seattleflu.org.  I updated the backoffice-automation IAM group to allow
PUTs to this new file and also setup a CloudFront distribution for the
bucket served at https://data.seattleflu.org.


kschwabefry (approved):

tsibley:
Deployed, but see #177.

PR #175, opened 2021-06-24
Include residuals from PHSKC in our UW EHR data requests

tsibley:
The lab does not currently record MRN and accession during specimen
intake, but the plan is to start soon.

### DIFF
--- a/bin/import-uw-retrospectives-to-redcap
+++ b/bin/import-uw-retrospectives-to-redcap
@@ -97,7 +97,7 @@ main() {
 
 parse-manifest() {
     id3c manifest parse-using-config "$CONFIG" \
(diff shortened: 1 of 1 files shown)


kschwabefry:
This will be great if we can the metadata for phskc residuals this way!

kschwabefry (approved):

tsibley:
I'm going to merge and deploy this, which will push a couple thousand existing samples into the REDCap project for requesting clinical metadata. However, these will not succeed in actually pulling metadata because the MRNs and accessions weren't recorded during sample intake. I don't believe this will cause any issues for the data pipeline (either our side or ITHS's).

PR #176, opened 2021-06-25
crontabs: Reduce HCT Redcap API batch size from 75 to 60

kschwabefry:
We are again getting 500 responses again from Redcap's API when we try
to pull records, using 75 records at a time for the HCT project.

I reproduced locally and then reduced to 60 and saw successful responses
from the Redcap API.


joverlee521 (approved):

kschwabefry:
Deployed.

PR #177, opened 2021-06-29
Run pathogen prevalence job under promjob

tsibley:
Oops, we want to capture metrics on all jobs!  This was an oversight on
my part.  (Another good reason to switch to systemd timers, where we can
more automatically and ~transparently capture these metrics.)


kschwabefry (approved):

tsibley:
Deployed.

PR #178, opened 2021-06-29
Run Shellcheck on every push as part of CI

tsibley:
Helps identify and avoid common shell pitfalls. Addresses all issues reported.


joverlee521 (approved):
I forget Shellcheck exists! Definitely useful to have, especially since I have minimal bash experience 😄

tsibley:
Merged and deployed.

PR #180, opened 2021-06-30
Quiet down the UW retros import

tsibley:
jq was throwing error messages about not being able to test() the
.sample_origin when it is null.

### DIFF
--- a/bin/import-uw-retrospectives-to-redcap
+++ b/bin/import-uw-retrospectives-to-redcap
@@ -101,7 +101,7 @@ main() {
 
 parse-manifest() {
     id3c manifest parse-using-config "$CONFIG" \
-        | jq --compact-output 'select(.sample_origin | test("^(uwmc|hmc|nwh|phskc)_retro$"; "i"))'
+        | jq --compact-output 'select(.sample_origin | strings | test("^(uwmc|hmc|nwh|phskc)_retro$"; "i"))'
 }
 
 diff-manifests() {


bin/import-uw-retrospectives-to-redcap
```diff
@@ -97,7 +97,7 @@ main() {
 
 parse-manifest() {
     id3c manifest parse-using-config "$CONFIG" \
-        | jq --compact-output 'select(.sample_origin | test("^(uwmc|hmc|nwh|phskc)_retro$"; "i"))'
+        | jq --compact-output 'select(.sample_origin | values | test("^(uwmc|hmc|nwh|phskc)_retro$"; "i"))'
```
joverlee521, line 100: Should we use `strings` instead of `values`? If I'm understanding the [docs](https://stedolan.github.io/jq/manual/#RegularexpressionsPCRE) correctly, `test()` only works for strings. We would run into similar errors if `.sample_origin` was a different non-null type.

tsibley: Ah, good catch. That is true:

```
jq> [4, "hello", "world", "!", null] | map(select(values | test("o")))
jq: error (at <unknown>): number (4) cannot be matched, as it is not a string

jq> [4, "hello", "world", "!", null] | map(select(strings | test("o")))
[
  "hello",
  "world"
]
```

Will fix.

joverlee521 (approved):

kschwabefry (approved):

tsibley:
Deployed.

PR #181, opened 2021-07-07
refresh-materialized-shipping-views: delete extra leading space

joverlee521:
The extra leading space in front of `__uw_encounters` started causing
errors due to the bash variable syntax change
in 74c8ead1a27bfa386148ab63d2e3dfb1b71b57a6.


kschwabefry:
Good catch!

PR #182, opened 2021-07-08
Publish weekly specimen counts every day

tsibley:
This data file will power a chart of specimen counts on seattleflu.org.


bin/export-prevalence
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+#
+# The samples included in this query should be kept in sync with
+# export-specimens.
```
kschwabefry, line 4: How important is it to keep these in sync? If we wanted to enforce this, I'm thinking we could combine export-prevalance and export-specimens into one bash script, do both initial queries, compare counts, and only upload to S3 if the check passes for both files.

kschwabefry (approved):

bin/export-prevalence
tsibley: It is not important enough to do that, IMO, which is why I left them separate and only added a comment. :-)

What the comment meant to convey (but maybe, as written, isn't doing well) is that the where clauses should remain ~identical, so that the two queries aggregate over the same set of samples.

If we wanted to ensure they're identical, I think the best approach would be to export a single, non-aggregated dataset of (anonymous sample id¹, collection week, site type, pathogen, present) and then use that to drive both charts, aggregating during charting. However, there's a little more to consider there from a data publication/privacy perspective² and so I punted on Just Doing It.

¹ Not necessarily `warehouse.sample.identifier`.
² One example: there might be instances of a small n for a given (collection week, site type).

tsibley: (We could also produce a single non-agged dataset that we then aggregate on in various ways and then publish, but again, punted on that for now.)

tsibley:
@kschwabefry Thinking some more, it might indeed be worth it to make the core shared query (below) a sqitch-managed view that both export jobs then use to produce the aggregates?

```sql
select *

from (
    select
        sample_id,
        to_char(coalesce(collected, encountered)::date, 'IYYY-"W"IW') as week,
        coalesce(site.details->>'type', 'unknown') as site_type,

        case
            when organism.lineage <@ 'Human_coronavirus.2019'
            then organism.lineage
            when organism.lineage <@ '{Human_parainfluenza, Human_coronavirus, Enterovirus}'::ltree[]
            then subpath(organism.lineage, 0, 1)
            else organism.lineage
        end as organism,

        present

    from warehouse.sample
    left join warehouse.encounter using (encounter_id)
    left join warehouse.site using (site_id)
    join warehouse.presence_absence using (sample_id)
    join warehouse.target using (target_id)
    join warehouse.organism using (organism_id)

    where
        (organism.lineage in
            ('Influenza.A.H1N1'
            ,'Influenza.A.H3N2'
            ,'Influenza.B'
            ,'RSV.A'
            ,'RSV.B'
            ,'Rhinovirus'
            ,'Adenovirus'
            ,'Human_metapneumovirus'
            ,'Human_coronavirus'
            ,'Human_parainfluenza'
            ) or organism.lineage <@ '{Human_parainfluenza, Human_coronavirus, Enterovirus}'::ltree[])
) as results

where
    week between '2018-W47' and to_char(current_date, 'IYYY-"W"IW')
```

tsibley:
Relies on https://github.com/seattleflu/id3c-customizations/pull/222/files.

tsibley:
Re-pushed with three changes:

- Mapping of sites → collection channels in the query. This makes the charting code a bit simpler, because no additional aggregation is necessary to account for multiple site types mapping to the same collection channel.
- Pre-calculate the weekly proportion of each channel in the dataset, so that this aggregate/window calculation can also be avoided in the charting code. Makes tooltip generation much simpler.
- Pull in site types for samples not attached to encounters via an existing view that matches against sample details. This reduces the "Other" channel by quite a bit across the whole time period, assigning them to the more specific channels. \o/

tsibley:
Deployed.

PR #183, opened 2021-07-09
return-of-results: use pagesize when fetching redcap records

kschwabefry:
Use the ID3C Redcap page_size param to fetch records in pages
instead of all at once for each Redcap project.

Fetching all the records at once was causing us to get 500 responses
on these requests with a message that "REDCap ran out of server memory"
responses from Redcap.

I'll start by setting page_size to 5000 for all the projects, which is
what we are currently using to fetch from HCT. We may need to tune it
later. This succeeded at the time of testing, but we will find out more
tonight, because we have more frequently been seeing the Redcap API
memory-related errors outside of business hours.

### DIFF
--- a/bin/return-of-results/export-redcap-scan
+++ b/bin/return-of-results/export-redcap-scan
@@ -58,7 +58,7 @@ def main():
 
 
 def fetch_records(project):
-    for record in project.records(fields = FIELDS, raw = True):
+    for record in project.records(fields = FIELDS, raw = True, page_size = 5000):
         if not all(len(record[field]) for field in REQUIRED_FIELDS):
             continue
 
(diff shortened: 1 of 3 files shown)


tsibley (approved):

joverlee521 (changes requested):
We should also add `page_size` to `fetch_encounter_records` within `export-redcap-sfs-longitudinal`, it seems to be the main source of the memory error in #id3c-alerts.

kschwabefry:
> We should also add `page_size` to `fetch_encounter_records` within `export-redcap-sfs-longitudinal`, it seems to be the main source of the memory error in #id3c-alerts.

Added! Thanks Jover.

joverlee521 (approved):

kschwabefry:
Deployed.

PR #185, opened 2021-07-14
terminate-idle-sessions: Exclude postgres user from consideration

tsibley:
This is the database "root" user, and should be excluded just like
"rdsadmin". We use "postgres" for `sqitch deploy`, which seems like it
can end up idle (or idle in a transaction) while it waits on a lock.


davereinhart (approved):

tsibley:
Deployed.

PR #186, opened 2021-07-14
etc/uw-retrospective-manifest: add aliquot sheet 15

davereinhart:
Adds aliquot sheet 15 to the list of worksheets for ingestion. Aliquot sheet 15
is the first and likely only worksheet that will be generated directly
from LIMS as an interim solution until we have an API developed accepting data in
real time.

This YAML file also switches to the new simplified format where multiple worksheets
sharing the same format can be entered as a list of values.

### DIFF
--- a/etc/uw-retrospectives-manifest.yaml
+++ b/etc/uw-retrospectives-manifest.yaml
@@ -1,5 +1,30 @@
 ---
-workbook: s3://fh-pi-bedford-t/seattleflu/bbi/2019_2020_sfs_aliquoting.xlsx
+workbooks:
+  - s3://fh-pi-bedford-t/seattleflu/bbi/2019_2020_sfs_aliquoting.xlsx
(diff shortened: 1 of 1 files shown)


etc/uw-retrospectives-manifest.yaml
```diff
@@ -238,3 +33,4 @@ extra_columns:
   mrn: mrn
   accession_no: accession
   sample_origin: sample_origin
+...
```
kschwabefry, line 36: I think this is a typo?

kschwabefry (changes requested):
Refactor and addition looks good, but I think an extra elipses snuck in at the end of the file.

etc/uw-retrospectives-manifest.yaml
davereinhart: I got this from YAML docs: 
https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics

kschwabefry (approved):

PR #188, opened 2021-07-15
Fix s3 protocol specification

seidelma:
Fix a bug with s3 protocol specification introduced because of differences in the way I tested locally.

### DIFF
--- a/bin/wa-doh-linelists/transform
+++ b/bin/wa-doh-linelists/transform
@@ -51,7 +51,7 @@ def put_file(linelist: pd.DataFrame, filename: str):
         filespec = fsspec.open(filename, mode='w',
             username=os.environ.get("DOH_USERNAME"),
             pkey=os.environ.get("DOH_PRIVKEY_PATH"))
-    elif (proto == 's3fs'):
+    elif (proto == 's3'):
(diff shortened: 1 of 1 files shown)


bin/wa-doh-linelists/transform
```diff
@@ -51,7 +51,7 @@ def put_file(linelist: pd.DataFrame, filename: str):
         filespec = fsspec.open(filename, mode='w',
             username=os.environ.get("DOH_USERNAME"),
             pkey=os.environ.get("DOH_PRIVKEY_PATH"))
-    elif (proto == 's3fs'):
```
kschwabefry, line 54: Question with your experience testing this - should we leave `s3fs` in addition to `s3`, for the sake of local testing, or do you think `s3` is sufficient here?

kschwabefry (approved):

bin/wa-doh-linelists/transform
seidelma: `s3` should be sufficient. I looked around the web earlier and `s3` is the correct protocol specification; however, fsspec.utils seems to just be capturing `(.*)://` and calling that the protocol. `ssh` & `sftp` are kind of unique in that either one is correct.

PR #189, opened 2021-07-16
Downgrade s3fs to 0.4.2 to fix linelist upload

seidelma:
s3fs 2021.06.1 doesn't seem to respect the AWS_DEFAULT_REGION environment variable, but it works in 0.4.2.
```
$ env|grep AWS
AWS_DEFAULT_REGION=us-west-2

$ pip list | grep s3fs
s3fs              2021.6.1
$ echo "import fsspec; fs=fsspec.open('s3://testing-bucket/afilethatexists.txt', mode='w')" | python3; echo $?
PermissionError: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.
1

$ pip list | grep s3fs
s3fs              0.4.2
$ echo "import fsspec; fs=fsspec.open('s3://testing-bucket/afilethatexists.txt', mode='w')" | python3; echo $?
0```

### DIFF
--- a/wa-doh-linelists/Pipfile
+++ b/wa-doh-linelists/Pipfile
@@ -8,7 +8,7 @@ verify_ssl = true
 [packages]
 id3c = {editable = true,git = "https://github.com/seattleflu/id3c.git",ref = "master"}
 seattleflu-id3c-customizations = {editable = true,git = "https://github.com/seattleflu/id3c-customizations.git",ref = "master"}
-s3fs = "*"
+s3fs = "==0.4.2"
 fsspec = "*"
 paramiko = "*"
 
--- a/wa-doh-linelists/Pipfile.lock
+++ b/wa-doh-linelists/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: wa-doh-linelists/Pipfile [hunk 1]
```diff
@@ -8,7 +8,7 @@ verify_ssl = true
 [packages]
 id3c = {editable = true,git = "https://github.com/seattleflu/id3c.git",ref = "master"}
 seattleflu-id3c-customizations = {editable = true,git = "https://github.com/seattleflu/id3c-customizations.git",ref = "master"}
-s3fs = "*"
+s3fs = "==0.4.2"
```
kschwabefry, line 11: Frustrating that newer versions of s3fs package keeps causing issues. We had to pin it for the backoffice id3c prod environment as well (https://github.com/seattleflu/backoffice/commit/92760f44ab4d794efe2437b03626fda7ad428ef9)

kschwabefry (approved):
Seems reasonable, thanks for digging in to get this working.
(Also, the description you put on the PR is really helpful, and would be great context to have in the git commit description as well!)

PR #192, opened 2021-07-26
Suppress alerts for samples with experimental collection barcode

davereinhart:
Update logging filter to suppress warnings for experimental samples
indicated by _exp suffix on the collection barcode. Previously the
filter was only suppressing warnings when the _exp suffix was present
on the sample barcode. With processing of tiny swabs there will
not be a sample barcode on the aliquoting sheet, only be a collection
barcode, so _exp suffix is being including there.


kschwabefry (approved):

davereinhart:
Deployed.

PR #194, opened 2021-08-02
Prevent overlapping instances of uw-ehs-report cron job

davereinhart:
The `uw-ehs-report` cron job is running hourly and regularly takes that
long to complete. Recent alerts regarding excessive duration of this task
(~20 hours) point toward issues related to when this job takes longer than
an hour to complete and runs again before the previous one has completed.

Adding flock to the command to prevent multiple instances from running, and
increasing the interval from 1 hour to 2 hours to give it adequate time to
finish before the next run.


crontabs/uw-ehs-report
```diff
@@ -13,6 +13,8 @@ ENVD=/opt/backoffice/id3c-production/env.d
 # Configure a private runtime directory for "fatigue".
 XDG_RUNTIME_DIR=/home/ubuntu/run
 
+# Directory to hold flock files
+FLOCK_DIR=/var/run/lock
 
 # Generate data file of results for UW Reopening (Husky Testing) and upload to EHS for followup
-7 * * * * ubuntu fatigue --quiet promjob "uw-ehs-report" pipenv run envdir $ENVD/redcap /opt/backoffice/bin/uw-ehs-report/generate
+7 */2 * * * ubuntu fatigue --quiet promjob "uw-ehs-report" flock --no-fork --nonblock "$FLOCK_DIR/uw-ehs-report" pipenv run envdir $ENVD/redcap /opt/backoffice/bin/uw-ehs-report/generate
```
kschwabefry, line 20: I might suggest supplying the additional flock param `--conflict-exit-code 0` here. If flock prevents a job from starting because the last one failed, it's probably ok to have it fail with an exit code of 0.

davereinhart: I saw that option and though about adding it, but with the interval increased to 2 hours flock shouldn't be triggered unless it starts twice as long as it currently does, so we may want to get an alert if it does happen.

davereinhart: `refresh materialized view` are the only cron jobs currently using that option, which I think we're ok with those failing silently but didn't think we would want the same for this one.

kschwabefry: I hear your point. I was originally thinking that it would typically be a no-op if we see that alert, and might wait til the existing run completes, but there may be cases where it would make sense to investigate what was happening, particularly if its for prolonged periods. (And it would give us an earlier notice than the grafana reduced job cadence warning).

kschwabefry (approved):

davereinhart:
Deployed

PR #195, opened 2021-08-10
Add new REDCap server HCT project to Return of Results

davereinhart:
New HCT REDCap server was built and project created for HCT 2021-2022.
Adding configuration for this server and project to Return of Results
to transition from existing ITHS server and HCT project hosted there.

Existing ITHS HCT project is also being left here until transition
to the new server is complete, at which point it can be removed from
this process.


davereinhart:
[HCT: Add new Redcap project to Return of Results](https://trello.com/c/l7e49Uq8/1152-hct-add-new-redcap-project-to-return-of-results)

kschwabefry (approved):
Looks right to me! I think all these fields have stayed the same from the ITHS redcap project, which is nice. 
I think I put this in trello too, but just echoing here as well- I'm thinking we don't want to deploy this yet, until that project moves to production mode and test data deleted, because we don't really want to accidentally push results for the test data in case they overlap with real barcodes.

davereinhart:
- [ ] requires environment variables `REDCAP_API_TOKEN_hct.redcap.rit.uw.edu_45` and `HCT_REDCAP_API_URL` which will be added with deployment of a PR in switchboard repo: https://github.com/seattleflu/switchboard/pull/56/files

davereinhart:
Deployed.

PR #198, opened 2021-08-16
Add support for multiple REDCap servers to wa-doh-linelists

seidelma:
For this ETL, each REDCap project ID is stored within the
wa-doh-linelists.yaml file. Move the REDCap API URL specification
to the file as well.

Using an environment variable was fine was fine when every data source
was on the same REDCap server. However, we're now breaking out the
HCT project onto its own REDCap server to handle the load. Since the
project ID is associated with a specific REDCAP API URL, we need to
preserve that association when more than one REDCap server is in use.

This seems like the most straightforward way to do that.


kschwabefry (commented):
Oh! I had forgotten that WADOH linelists for HCT use the EHS transfer project instead of the core HCT project, so these will all be staying with the same server URL for now. We will still want the multi-server flexibility once we switch that EHS project over later as well.

PR #90, opened 2020-10-15
Add SFS Childcare to ROR

joverlee521:
* Generalize the UW-reopening REDCap export script to export any SFS longitudinal REDCap project that is defined in the `etc/sfs-longitudinal-redcap-projects.yaml` config file.

* Update the ROR transform script to required contacted field for SFS Childcare studies

* Check for duplicate barcodes across __all__ studies instead of just within each study arm.

PR #199, opened 2021-08-16
Filter REDCAP query to only pull records required for uw-ehs-report job

davereinhart:
The uw-ehs-report job only needs specific records from REDCap for joining
to ID3C data. This adds some additional filter to the REDCap request to only
pull records from the relevant requests and for the repeating encounter event
instruments, an additional filter is applied to only select completed forms
of the type needed.

This results in a significant reduction in the amount of data being requested from
REDCap API and smaller resulting ndjson file (less than 1/10th the size), which in turn
leads to faster processing of the transform script where pandas dataframe is used
to join REDCap and  ID3C records.


bin/uw-ehs-report/transform
```diff
@@ -85,7 +85,7 @@ def find_duplicate_netids(redcap_file: str) -> List[str]:
     with open(redcap_file, 'r') as f:
         for line in f:
             line_json = json.loads(line)
-            if line_json['netid']: # Only enrollment records have netid populated
+            if line_json.get('netid'): # Only enrollment records have netid populated
 
                 if line_json['netid'] in found_netids:
```
seidelma, line 90: Will this throw a KeyError if `line_json['netid']` is undefined, as on line 88?

seidelma (approved):
Looks good to me, just one question about the transform script.

bin/uw-ehs-report/transform
davereinhart: It shouldn't get to line #90 if `netid` key is missing, since `line_json.get('netid')` will return `None`

davereinhart:
I did a little more testing locally of this, comparing final result with the old method using diff, and there are 4 additional rows (out of 98765) in the transform script's resulting ndjson file using this branch. Still investigating what's unique about those records that could be causing that.

davereinhart:
> I did a little more testing locally of this, comparing final result with the old method using diff, and there are 4 additional rows (out of 98765) in the transform script's resulting ndjson file using this branch. Still investigating what's unique about those records that could be causing that.

Confirmed... the current version of `export-redcap-uw-reopening` in master branch is missing at least 4 "redcap_repeat_instance" records that should be in the `redcap_data.ndjson`, and that are included when using the updated version of that script in this branch. This results in 4 additional records in the final output of the `transform` script. I found no reason those should have been excluded, and in fact those 4 barcodes do appear in the "EH&S Data Transfer for Husky Coronavirus Testing" REDCap project where the results are exported, so they must have been included in previous runs. The large number of records being requested from REDCap may have something to do with it (which is the main issue this branch is targeting), but I would have expected an error rather than instrument instances being dropped at random.

Going to move forward with merging and deploying this branch despite this very small (0.004%) and unexplained difference in results, since the results with this change look more accurate than what is currently deployed.

bin/uw-ehs-report/export-redcap-uw-reopening
```diff
@@ -11,45 +11,55 @@ from typing import NamedTuple, Optional
...
-]
+            # barcodes
+            "collect_barcode_kiosk",
+            "return_utm_barcode",
 
+            # non PII fields that we don't ingest into ID3C
+            "illness_kiosk",
+            "illness_swabsend",
+        ],
+        "filter":"[kiosk_registration_4c7f_complete]=2 or [husky_test_kit_registration_complete]=2 or [post_collection_data_entry_qc_complete]=2"
```
kschwabefry, line 60: It might be helpful to add a comment here to explain why you're filtering to these instruments being complete, something like:

Filter the redcap export to record instances in which we'd expect to be able to match to a sample, which involves having a completed kiosk registration, test kit registration, or PCDEQC form.

kschwabefry (approved):
Let's keep debugging the issue with the four extra instances showing up in master branch and not this, but I suspect that might be an issue with how we are doing paging in redcap.py module, not this PR.

Thanks for these changes, this will be a nice improvement.

bin/uw-ehs-report/export-redcap-uw-reopening
davereinhart: Good suggestion!  Comment added.

bin/uw-ehs-report/transform
davereinhart: @seidelma I cleaned up the code a little here. It was a little messy, and just needed to retrieve the `netid` once.

davereinhart:
Deployed.

PR #200, opened 2021-08-24
etc: include kit reg barcode in HCT results

kschwabefry:
Include the redcap field barcode_swabsend for the HCT projct when
exporting redcap data to use for return of results.
The new samples with tiny swabs won't have PCDEQC forms filled out,
so we will need to use barcode_swabsend to link the results data to
a redcap record.


davereinhart (approved):
Looks good to me! I ran it locally and confirmed the result does include kit registration barcodes.

kschwabefry:
Deployed.

PR #204, opened 2021-08-27
Add more logging around duplicate REDCap records

seidelma:
Refactor the drop_duplicate_barcodes() function to also return a list of the barcodes that were duplicated (or None if none were found).

When a duplicate barcode is found in a REDCap data frame, also output the record_ids that contain it, to aid in tracking down the errors that caused them, as well as where in the process the duplicate barcode was found. It turns out that record_ids can also be duplicated, so also deduplicate them before printing where we found them and whether any were repeated.


seidelma:
Refactored those code blocks into a separate function as @davereinhart suggested.

davereinhart (approved):
Works well!  

Not a blocker, but one possible way get pull the duplicate barcode records more efficiently from the dataframe would be to use the duplicates (`dups`) index values to select the relevant records from the full set of records (`redcap_data`):
```
df_dups = redcap_data[['record_id','prioritized_barcode']][redcap_data.index.isin(dups.index)].drop_duplicates()
```
Then `groupby` and `agg` can be used to summarize down to a list of records ids for each duplicated barcode:
```
df_dups = df_dups.groupby(['prioritized_barcode']).agg({'record_id': ', '.join})
```

davereinhart:
@seidelma One other thing worth noting, `prioritized_barcode` isn't a field in REDCap, it's only a field in the dataframe, so the error message won't indicate exactly where in REDCap to find each duplicated barcode for the non-SCAN records.

PR #205, opened 2021-09-01
crontab: increase run frequency of backfill-sample-ids

kschwabefry:
The reportable_conditions job only will report for samples
that have a sample id. The tiny swabs samples do not have a
sample id, until the backfill-sample-ids job runs, and we want
them included in the reportable conditions notifications.

So, we will increase the frequency at which we are assigning the
sample ids, so that we can trigger the reportable conditions notifications
on a more timely basis.

Note - I left this running once an hour for now. This morning it took 18 minutes to run,
but we'll be able to gather more information about runtime as we go.


davereinhart (approved):

kschwabefry:
Deployed.

PR #209, opened 2021-09-08
Check for no encounters in project

davereinhart:
Checking for records in encounters dataframe before additional processing and
joining with enrollment dataframe.


davereinhart:
Deployed.

PR #213, opened 2021-09-10
Removing duplicate barcodes before import to REDCap

davereinhart:
Adds an additional step to the process to sort and remove duplicates barcode
lines from the ndjson file before running the `import-to-uw-ehs-transfer-redcap`
script. Previously, duplicate barcodes were making it through the process
because the transform check only checks for duplicates within each batch of
10K records.

Due to memory limitations we can't load the entire file at once to
check for duplicates in a dataframe, so this branch handles the check by sorting
looping throught the file line by line, first compiling an array of all the duplicate
barcodes, then looping through it a second time to skip the lines containing those barcodes.


davereinhart:
Deployed.

davereinhart:
[Fix EHS duplicates coming through](https://trello.com/c/O3FPkt65/1198-fix-ehs-duplicates-coming-through)

PR #214, opened 2021-09-11
initialize last_barcode variable with empty string

davereinhart:


davereinhart:
[Fix EHS duplicates coming through](https://trello.com/c/O3FPkt65/1198-fix-ehs-duplicates-coming-through)

PR #217, opened 2021-09-14
Return-of-results remove barcodes not testable

davereinhart:
Removes barcodes marked as not testable before removing duplicate barcodes. This ensures these samples will not be included in return of results even if they are processed by the lab, and reduces the number of duplicate barcodes that show up as warnings and actually do need to be resolved.


bin/return-of-results/transform
```diff
@@ -83,12 +97,16 @@ def parse_scan_redcap(redcap_file) -> pd.DataFrame:
     kiosk_barcodes = utm_tube_barcode_2.combine_first(reenter_barcode)
 
     barcodes = kiosk_barcodes.where(collection_via == "kiosk", mail_barcodes)
+    
+    # Drop barcodes marked as not testable
+    able_to_test = redcap_data['able_to_test']
+    barcodes = drop_not_testable(barcodes, able_to_test, 'SCAN')
```
kschwabefry, line 103: Just had a chat with Melissa to verify, sounds like this does make sense from lab perspective to drop these before dropping duplicates, in these cases where there are duplicates and one was testable and one wasn't, its ok/desired to return the testable one, even if there's a duplicate barcode. 
✔️

kschwabefry (approved):

bin/return-of-results/transform
davereinhart: Ok, thanks for checking on that!

davereinhart:
Deployed.

PR #218, opened 2021-09-14
Switch back to using print instead of log.warning

davereinhart:
Switching back to using print statements instead of log.warning so that
warnings don't include a timestamp. The timestamp was causing the warnings to
not be caught by fatigue, since they were different each time the job runs.


kschwabefry (approved):

davereinhart:
Deployed.

PR #219, opened 2021-09-15
include record_id when returning parsed redcap records

kschwabefry:
This is a hotfix for an issue we are seeing with the pipeline, where the dataset didn't have record_id when its looking at cross-project duplicates, which was causing the process to error out when it tried to log the record_id for those duplicates.


kschwabefry:
deployed.

PR #220, opened 2021-09-16
Reverting prior change that dropped non-testable records

davereinhart:
Additional context was provided that samples marked as not testable
should still receive return of results ("Unable to test" is one of the
possible status codes on the result returned to the participant).

This reverts the privious change that was dropped those record in an
attempt to reduce the number of duplicate barcode warnings being received.
Discussions are ongoing with the study team on how to handle duplicate
barcodes and whether or not results with status "Unable to test" should be
returned for those records, filtered out during return of results if they
are marked as not testable, or if the duplicate barcodes should be removed
from REDCap.


kschwabefry (approved):

davereinhart:
Deployed.

PR #221, opened 2021-09-20
Ror duplicate barcode handling

davereinhart:


davereinhart:
[Filter out duplicate barcode alerts](https://trello.com/c/lkhkdqSD/1199-filter-out-duplicate-barcode-alerts)

davereinhart:
`[able_to_test]="no"` in REDCap PCDEQC form should correspond to `status_code = 'never-tested'` in ID3C `return_results_v3` view for REDCap projects that have scheduled ETL jobs. There will be a delay in those values being updated in ID3C, so although adding a confirmation check at the end of the RoR process would be ideal it would need to be ignored for certain projects and would need to allow a window for ETLs to run before skipping records or throwing warnings for those values not matching. Currently weighing options for how to handle that, or whether to just skip checking REDCap `able_to_test` value and only rely on ID3C `status_code`.

PR #224, opened 2021-09-23
Add xls/xlsx to allowed file extensions for year 1

seidelma:
Fix an issue preventing the retrospective from running because a file was uploaded with an extension we don't recognize.


kschwabefry (approved):
Sounds good to me. So, for the year 1 files then we'll just leave them in S3 without processing, until we make further updates to handle them. Thanks Todd!

seidelma:
Functionality already merged.

PR #226, opened 2021-09-28
Fix argument consumption in refresh-database

seidelma:
Fix `--exclude-receiving-data` when used with a specific tmpdir argument.


seidelma:
Fixed as part of https://github.com/seattleflu/backoffice/pull/238.

PR #227, opened 2021-10-01
Make uw-ehs-report noisier if it fails to acquire the lockfile

seidelma:
Move the lock acquisition down into the uw-ehs-report script
so we can be more obvious when it fails. Previously, the cron job
would just fail with nonzero status if it couldn't acquire the
lock, and no message was sent to alerting channels. Since this
report is a priority, we want to know about it ASAP.


seidelma:
This was addressed in a better way by #240. Closing!

PR #232, opened 2021-10-20
Replace REDCap URL with hostname for alert formatting

davereinhart:
Redo of https://github.com/seattleflu/backoffice/pull/221 which I merged, deployed, then reverted.

With one modification:
Including a URL in alert to Slack was causing those to be translated
into sendgrid.net URLs. Simplifying here to just the REDCap
hostname since that is all we need for these alerts.


bin/return-of-results/export-redcap-scan
```diff
@@ -59,13 +60,14 @@ def main():
...
 def fetch_records(project):
+    redcap_host = Url(project.base_url).netloc
     for record in project.records(fields = FIELDS, raw = True, page_size = 5000):
         if not all(len(record[field]) for field in REQUIRED_FIELDS):
             continue
 
         yield {
             "project_collection_via": project.collection_via,
-            "redcap_url": project.base_url,
+            "redcap_host": redcap_host,
```
kschwabefry, line 70: Can you share an example of what the new string will look like? I'm curious if sendgrid will still try to replace it.

davereinhart: hct.redcap.rit.uw.edu

kschwabefry: Hmm, I might expect that to be rewritten by sendgrid as well, I had thought it scanned text for things that looked like URLs. Have you found a way to test it out?

davereinhart: I haven't tested with sendgrid, but I can look into that.

davereinhart:
Deployed.

PR #238, opened 2021-11-12
refresh-database fixes & improvements

seidelma:
Add checks for the work directory and target database prerequisites,
so we don't accidentally download the entire production database
and then fail when something required for successful completion
isn't installed or whatever. Again.

Also clean up option parsing, so it doesn't matter
which order you put options in.


seidelma:
Force-pushing just cleaned up some spaces and trivial changes that didn't need to be diffed

kschwabefry (approved):

PR #240, opened 2021-11-29
Fatigue: Log error message and fail if lock cannot be acquired

davereinhart:
Currently, if fatigue cannot acquire a lock file it fails silently. This change will exit with an error code and log a message in that scenario, so that the failed job will generate an alert and appear on Grafana dashboards.

Also a cron job is being added to clear the fatigue folder on reboot to help ensure no lock files persist. We have recently had cases of jobs failing after system reboot, and having fatigue start with a clean slate will help eliminate that as a possible cause.


davereinhart:
To test this change out, I spun up a VM using vagrant running Ubuntu (for flock requirement), then copied the fatigue script over. There are two environment variables to set: XDG_RUNTIME_DIR can be set to your home folder and FATIGUE_LOCK_TIMEOUT's default is 60 seconds but setting it to 5 makes it easier to test here.

Run fatigue on a command as background process (`sleep 20` is the command we're applying it to):
`XDG_RUNTIME_DIR=/home/vagrant FATIGUE_LOCK_TIMEOUT=5 ./bin/fatigue sleep 20 &`

Then immediately run fatigue in the foreground on the same command:
`XDG_RUNTIME_DIR=/home/vagrant FATIGUE_LOCK_TIMEOUT=5 ./bin/fatigue sleep 20`

You should see the new error message.

davereinhart:
[RoR Change - Make alert more visible / Make Fatigue noisier](https://trello.com/c/vvoPKY3N/1222-ror-change-make-alert-more-visible-make-fatigue-noisier)

kschwabefry (approved):

davereinhart:
Deployed. (with additional fix in PR #244)

PR #242, opened 2021-12-01
Reducing batch size to prevent request timeout

davereinhart:
After implementing a 300-second client-side timeout on REDCap requests, the uw-ehs-report
job began failing when posting data back to REDCap. The REDCap server-side timeout has
been communicated as being set to 120 seconds. To help keep requests under both timeout limits,
the batch size is being reduced from 8500 to 2500.


kschwabefry (approved):

davereinhart:
Deployed.

PR #246, opened 2021-12-06
huksy-musher uWSGI configuration update

davereinhart:
Updating husky-musher uwsgi configuration based to be more stable in
production. More information on the each of these settings is availabe in
the uwsgi documenation.

These changes are specifically targeted at the recurring issue of husky-musher
worker threads getting stuck in a read state, which may be the result of
forking a multi-threaded application (lazy-apps will load application
one time per worker) and/or the "thundering herd" problem described in
the uwsgi documentation (thunder-lock should be enabled by default
according the article in the docs, but it isn't).

The other settings (strict, vacuum, harakiri, die-on-term, max-requests,
need-app, single-interpreter) are generally recommended in production.


davereinhart:
Documentation and additional resources:
https://uwsgi-docs-additions.readthedocs.io/en/latest/Options.html
https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/
https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html#the-zeeg-problem-multiple-processes-with-multiple-threads
https://shareablee.github.io/blog/stalled-uwsgi-workers/

davereinhart:
[Investigate EC2 restart and stuck cron jobs](https://trello.com/c/2rPaXPwU/1247-investigate-ec2-restart-and-stuck-cron-jobs)

davereinhart:
[Update musher to not use participant data from UW's IDP](https://trello.com/c/weecGrSV/1268-update-musher-to-not-use-participant-data-from-uws-idp)

kschwabefry:
Do you have thoughts on how we can test the config changes?

davereinhart:
Maybe the virtual box that @seidelma is working on could be used for testing the config changes?

I planned on deploying this change after hours and monitoring it closely to roll back if needed.

To see if it has the desired effect, I added a new panel in Grafana that shows the total # of requests per worker. Currently they are not very evenly distributed, but after the change that should improve:
https://grafana.seattleflu.org/d/ZOdQQGEMz/web?from=now-90d&to=now&orgId=1&viewPanel=16

kschwabefry (approved):
Thanks for adding that new monitoring panel. I've read through the changes and docs here, and it seems reasonable - although hard to know for sure how it will impact behavior without seeing it play out.

@seidelma can you take a read through this config change as well?

davereinhart:
Deployed.

PR #247, opened 2021-12-10
etc: use batlab file prefix for aliquot sheets

kschwabefry:
We just moved these files to a new S3 bucket, and changed the prefix
to from bbi to batlab to be a little more meaningful for our users.


kschwabefry:
[Transition FH S3 bucket to SFS PHI Bucket](https://trello.com/c/LJIwCTP4/1175-transition-fh-s3-bucket-to-sfs-phi-bucket)

davereinhart (approved):

kschwabefry:
Deployed.

PR #251, opened 2021-12-28
Changing cadence of uw-ehs-report job to hourly

davereinhart:
Alerts have been more common as this job frequently takes over 30 minutes
to complete. Changing to run hourly instead of every 30 minutes to prevent
overlapping runs and related alerts related to lock files.


davereinhart:
Deployed.

PR #253, opened 2022-01-03
[Do not merge] Changes required for Python 3.9

seidelma:
Don't merge this (yet)--these are the changes required to get the backoffice repo to work with Python 3.9. They may be backward-incompatible; I'm just figuring out what is required to make Python 3.9 work first.


seidelma:
There's something of a gotcha here: the `Pipfile`s specify that the id3c* packages should use the `master` branch, but that branch doesn't have the Python 3.9 fixes, so we've got to change it to use my `seidelma/bugfix/python-3.9` branches. I'm not committing that `Pipfile` change back to git, because we want to preserve the branch name in there. 

So if you want to test it with the updated package requirements, locally change `ref = "master"` to `ref = "seidelma/bugfix/python-3.9"` for the `id3c` and `seattleflu-id3c-customizations` packages in each `Pipfile` that uses them.

davereinhart:
@seidelma It may be worthwhile to add some of the additional setup steps to `dev/vagrant/backoffice/provision.sh`, like cloning the backoffice repo, assigning permissions and creating users (i.e. `ubuntu`, `www-data`), and setting up python 3.9.9, just to make sure we're testing with the same environment.

Also, is the Vagrant VM is just being used to test the cron jobs, not the web apps or API?

PR #254, opened 2022-01-05
Add new environment variables for REDCap server status

davereinhart:
Two boolean environment variables are being added to indicate the expected status
of the two REDCap servers. True will indicate the server is expected to be online.
These variables will be used to prevent scheduled tasks from running when there
is an expected outage, and have been added to the appropriate cron jobs.

For now, these will need to be toggled manually when there is a scheduled outage.
A possible future improvement would be to toggle these automatically based on the
results of a health check on each server. A single alert would be needed if the status
changed. This would prevent failing jobs and alerts we typically see during an unexpected
REDCap outage.


kschwabefry (approved):
LGTM. One thing I see is that for commands running with `pipenv`, we bounce around a bit between putting this before and after invoking `pipenv`, but I don't /think/ that will have a functional impact 🤔 and there are preexisting inconsistencies with whether `envdir` is invoked before/after `pipenv run`.

It would be great to follow this up with an update to the [troubleshooting docs](https://github.com/seattleflu/documentation/blob/master/Troubleshooting.md)

davereinhart:
> LGTM. One thing I see is that for commands running with `pipenv`, we bounce around a bit between putting this before and after invoking `pipenv`, but I don't /think/ that will have a functional impact 🤔 and there are preexisting inconsistencies with whether `envdir` is invoked before/after `pipenv run`.
> 
> It would be great to follow this up with an update to the [troubleshooting docs](https://github.com/seattleflu/documentation/blob/master/Troubleshooting.md)

Agreed, I've noticed that too. When I run these commands locally, I always put the `pipenv` command after all `envdir`s, so it would be good to make that consistent here, but it doesn't seem to be causing any problems.

davereinhart:
Deployed.

PR #255, opened 2022-01-05
Rename Internal SCAN Switchboard Mentions to SFS Switchboard

bencap:
The SCAN switchboard is now being called the SFS Switchboard, and external facing parts of this switchboard have been renamed (see [this PR](https://github.com/seattleflu/switchboard/pull/22)). This change renames switchboard directories, service files, and crontabs to match the name change.

sfs-switchboard has been created as a copy of scan-switchboard, so references to this new directory should work once deployed, at which point scan-switchboard can be removed.


bencap:
[Rename internal mentions of "SCAN switchboard"](https://trello.com/c/hpw5yZOa/832-rename-internal-mentions-of-scan-switchboard)

crontabs/sfs-switchboard
```diff
@@ -0,0 +1,12 @@
+SHELL=/bin/bash
```
kschwabefry, line 1: Looks good! I think there's another PR that modifies these same files https://github.com/seattleflu/backoffice/pull/258 so it will be worth making sure those same changes get picked up when this is merged and deployed

kschwabefry (commented):
Thanks for making these changes really thoroughly, looks like we caught it in many different places. 

One thing is that the in the commit history, it looks like some commits from master were pulled into this branch with different commit ids. Probably an artifact of local rebases or merges, but we should take another look at the git usage and workflows to make sure we're following the same practices.

bencap:
@kschwabefry yeah so what happened I think is that Dave pushed the changes for his environment variables, and then I tried to rebase this branch off of that environment variable one. I will go back through and see about cleaning up the history - i am just not super experienced with other git commands outside of commit/push/pull

kschwabefry:
> @kschwabefry yeah so what happened I think is that Dave pushed the changes for his environment variables, and then I tried to rebase this branch off of that environment variable one. I will go back through and see about cleaning up the history - i am just not super experienced with other git commands outside of commit/push/pull

No worries at all, lets go over during 1:1 today. Git can take a little practice sometimes and different teams use it differently!

kschwabefry (approved):

bencap:
deployed

PR #256, opened 2022-01-05
Remove Hold on SCAN RoR

truong97:
Update the hardcoded piece in transform script to remove positive and inconclusive results hold for SCAN as approved the the IRB.

The end goal is to refactor SCAN RoR into a config file; however due to high priority and time restraint, this will be created in a later PR.


truong97:
[Remove positive RoR hold for SCAN samples](https://trello.com/c/uZ8GM0pq/1288-remove-positive-ror-hold-for-scan-samples)

bin/return-of-results/transform
```diff
@@ -173,7 +173,7 @@ def edit_status_code(results: pd.DataFrame) -> pd.DataFrame:
     # To prevent a row's `require_contacted` value from getting set as pd.NA and then failing where we compare
     # its value with the boolean values of the other series (is_positive, is_inconclusive, participant_contacted),
     # first check that the value is not pd.NA before comparing its string value.
-    require_contacted_lambda = lambda row: (row['source'] is not pd.NA and row['source'] == 'scan') \
+    require_contacted_lambda = lambda row: (row['source'] is not pd.NA) \
```
kschwabefry, line 176: Let's take another look at this logic here.

Before, the logic was:
```(if the row has a non-NA source, and that source is SCAN) or (if the row has a non-NA contact_before_releasing_result and that contact_before_releasing_result is true) ```

With this change, the logic would be:
```(if the row has a non-NA source) or (if the row has a non-NA contact_before_releasing_result and that contact_before_releasing_result is true) ```

This would result in anything with a non-NA source having `require_contacted_lambda` set to true, which isn't what we want here, because that would mean that any project with a defined source, we'd always write-over positive results with pending, until that pt is contacted. 

Try changing this whole line to:
```require_contacted_lambda = lambda row: (row['contact_before_releasing_result'] is not pd.NA and row['contact_before_releasing_result'] == 'true')```

truong97: AHHHH I totally overlooked that! Thanks for feedback, I went in and modified it to the suggested line above.

truong97:
This PR is ready for review. Testing has been done and confirmed that the code changes produces the expected results (unholding positive/inconclusive SCAN results). 

Method of testing include:
1) outputting json file with branch - running generate results csv
2) outputting json file with master - running generate results csv
3) comparing the two files via diff
4) comparing the diff to a REDCap report with pending SCAN RoR
5) check status on secure link

kschwabefry (approved):

bencap:
deployed

PR #258, opened 2022-01-06
Add new environment variables for REDCap server status

davereinhart:
Two environment variable files are being added to indicate the whether or not the two
REDCap servers are online or not. These presence of `REDCAP_OFFLINE` or `HCT_REDCAP_OFFLINE` files
in `env.d/redcap`  indicates the corresponding server is expected to be offline. If the file is
missing, the server is expected to be online. This can be used to prevent cron jobs that utilize
REDCap from running when there is an scheduled outage.

For now, these will need to be created and removed manually when there is a scheduled outage.
A possible future improvement would be to toggle these automatically based on the results of a
health check on each server.


crontabs/return-of-results
```diff
@@ -15,4 +15,4 @@ XDG_RUNTIME_DIR=/home/ubuntu/run
 
 
 # Generate data file and PDFs for returning SFS results via UW Lab Med's SecureLink portal
-5,35 * * * * ubuntu fatigue --quiet promjob "return of results" flock --no-fork --nonblock /var/lock/return-of-results pipenv run envdir $ENVD/redcap envdir $ENVD/securelink/ /opt/backoffice/bin/return-of-results/generate
+5,35 * * * * ubuntu fatigue --quiet promjob "return of results" flock --no-fork --nonblock /var/lock/return-of-results pipenv run envdir $ENVD/redcap envdir $ENVD/securelink/ REDCAP_ONLINE && HCT_REDCAP_ONLINE && /opt/backoffice/bin/return-of-results/generate
```
kschwabefry, line 18: Assuming it looks good with local testing, should the same syntax change to `[ -n $REDCAP_ONLINE ]` be applied to this and the remaining crontabs too?

davereinhart: Actually, that doesn't seem to be working either. I'm not sure we can reference the environment variables here when they're coming from `envdir`, it looks like they might only available from within the command that's running.

davereinhart: Since we can't use `envdir` here, these environment variables can either be added at system level or at the top of each cron file.  Let me think about it some more, I'll update this PR shortly.

davereinhart: I updated the PR and ended up going with @seidelma's suggestion of presence/absence of a file, so to pause the cron jobs we just need to delete one or both of the new files I added to `env.d/redcap`  I also moved the checks to the very beginning of each command since we're no longer relying on `envdir` to load those.

kschwabefry (approved):

davereinhart:
[Pause cron jobs during planned REDCap outage using environment variable](https://trello.com/c/Loblz7O1/1287-pause-cron-jobs-during-planned-redcap-outage-using-environment-variable)

PR #260, opened 2022-01-10
Remove Dropped Study Fields

bencap:
This change removes references to the `study_tier` and `is_uw_greek_member` fields, which were recently removed and eliminated from the database view.


kschwabefry (commented):
The changes look good for this script. I'm curious how redcap will behave when this is deployed. Since it drops these variables, when we post the dataset, I wonder whether it would consider existing records to be updated. We might want to test with a record or two. If it considers them as updates that might cause unnecessary work in redcap, and potentially duplicate redcap alerts, depending how they are configured.

We could potentially side step this by coordinating this deployment with the deletion of the no-longer used variables from the EHS redcap project - although before we do that we might want to talk with the EHS team to see if they were using that data historically (and maybe Jess on the HCT study team to check her alerts currently set up - I spot checked and saw is_uw_greek_member being used in one of the configured redcap alerts.

kschwabefry (approved):

bencap:
deployed

PR #261, opened 2022-01-14
Add scan-switchboard to Old Crontabs List

bencap:
When I had previously deployed the name change to `scan-switchboard`, I deployed it as if it were an existing application rather than acting as if it was a new one (although functionally the same as the old crontab, it is 'new' because it has a new name). This change will remove the old `scan-switchboard` crontab that is still running when the `sfs-switchboard` crontab gets installed.


kschwabefry (approved):

bencap:
Deployed

PR #263, opened 2022-01-22
Quiet Alerts for HCT Duplicate Barcodes

bencap:
With the HCT study team responsible for tracking duplicate barcodes in HCT, the duplicate alerts for the HCT study are no longer actionable and clog up the id3c-alerts channel. This change silences ROR duplicate barcode alerts when all duplicate barcodes are part of the HCT study (if duplicates are in other studies, or if they are duplicated across HCT and others, they will still be pushed to id3c-alerts). This change also moves duplicate barcodes in the EH&S job to info level logging instead of warning level. Duplicate netids in the EH&S job remain at warning level.


bin/return-of-results/transform
```diff
@@ -234,18 +234,23 @@ if __name__ == '__main__':
...
-              f"from Return of Results: {duplicate_barcodes} ", file=sys.stderr)
-        print(f"{duplicates[list_of_cols].to_string(index=False)}", file=sys.stderr)
+        hct_barcodes = [barcode for barcode in duplicate_barcodes if
+            (duplicates.loc[duplicates['qrcode'] == barcode]['project_id'] == '45').all()]
+
+        # don't print warning if all the duplicate barcodes are from the hct study
+        if len(duplicate_barcodes) != len(hct_barcodes):
+            print(f"Dropping {len(duplicates)} rows containing {len(duplicate_barcodes)} duplicated barcodes "
+                f"({len(hct_barcodes)} HCT only barcodes) from Return of Results: {duplicate_barcodes} ", file=sys.stderr)
+            print(f"{duplicates[list_of_cols].to_string(index=False)}", file=sys.stderr)
```
davereinhart, line 251: Rather than checking to see if all of the duplicate barcodes are from HCT, we can just filter the list to only print those that are non-HCT, which should be doable with a dataframe query.

bin/uw-ehs-report/transform
```diff
@@ -210,6 +210,9 @@ def process_redcap_batch_and_dump_output(rows: List[dict], id3c_data: pd.DataFra
 
         # find and drop rows that contain duplicate netids
         found_duplicate_netids = enrollments.reset_index().merge(duplicate_netids, how='inner', on=['redcap_url', 'project_id', 'record_id']).set_index('index')
+        if len(found_duplicate_netids) > 0:
+            log.warn(f"Dropped {len(duplicate_netids):,} records with duplicated netids: <masked>")
+
```
davereinhart, line 215: Is this additional logging needed, or will this same message be printed in the `drop_duplicate_values` function?

davereinhart (changes requested):

bin/uw-ehs-report/transform
bencap: Drop duplicate values is not used when dropping netids, but you are right that this is unnecessary.

bin/return-of-results/transform
bencap: Thanks for the tips on this, new changes fix this!

bencap:
Made these changes and tried my hand at rebasing, for when you review again @davereinhart did I do that how I am supposed to?

bin/return-of-results/transform
```diff
@@ -234,18 +234,24 @@ if __name__ == '__main__':
...
-    
+
     # Drop any duplicate barcodes
     duplicates = redcap_data[redcap_data['duplicate']].sort_values('qrcode')
+    non_hct_duplicates = duplicates['qrcode'][duplicates['project_id'] != '45'].drop_duplicates()
+    duplicates_alerting = duplicates.merge(non_hct_duplicates, how='inner', on='qrcode')
+
+    # only alert if the barcodes are not all hct
+    if not duplicates_alerting.empty:
+        duplicate_barcodes = list(duplicates_alerting['qrcode'].unique())
```
davereinhart, line 247: Not a blocker, but I think you already have these values in `non_hct_duplicates` so you should be able to use that rather than calling `unique()` on `duplciates_alerting`

bencap: Ah true nice eye. Seems more logical to use that when I am explicitly printing non-HCT barcodes. Will change that rn.

bencap:
[Quiet duplicate barcode alerts for HCT](https://trello.com/c/PD7T69om/1271-quiet-duplicate-barcode-alerts-for-hct)

davereinhart (approved):

bencap:
deployed

PR #264, opened 2022-01-24
Temporarily pause cronjobs for Postgres upgrade

seidelma:
Temporarily pause cronjobs from approximately 3am to 5am so the database upgrade can proceed without jobs throwing errors because they can't connect.

Some jobs start real a little too close to the 3am maintenance window opening (like 2:55) for comfort, so don't run then, either. It should be safe to start running jobs again around 5am.


crontabs/id3c-production
```diff
@@ -99,25 +99,25 @@ GEOCODING_CACHE=/home/ubuntu/scan-cache.pickle
...
 # Use redcap-api-batch-size = 1000 because this REDCap project is longitudinal.
 GEOCODING_CACHE=/home/ubuntu/sfs-cache.pickle
-21,51 * * * * ubuntu promjob "id3c etl redcap-det childcare-2021" flock -F $GEOCODING_CACHE envdir $ENVD/deidentification envdir $ENVD/smartystreets envdir $ENVD/redcap pipenv run id3c etl redcap-det childcare-2021 --redcap-api-batch-size 1000 --commit
+21,51 0-1,5-23 * * * ubuntu promjob "id3c etl redcap-det childcare-2021" flock -F $GEOCODING_CACHE envdir $ENVD/deidentification envdir $ENVD/smartystreets envdir $ENVD/redcap pipenv run id3c etl redcap-det childcare-2021 --redcap-api-batch-size 1000 --commit
 
 # Ingest FHIR documents every 5 minutes.
-*/5  * * * * ubuntu fatigue --quiet promjob "id3c etl fhir" pipenv run id3c etl fhir --commit
+*/5  0-2,5-23 * * * ubuntu fatigue --quiet promjob "id3c etl fhir" pipenv run id3c etl fhir --commit
 
 # Run the idle database session disconnector routine every minute
```
kschwabefry, line 122: I don't expect this to be a issue for tomorrow, but for thoroughness, next time we make an upgrade we might also pause these for good measure.

kschwabefry (approved):
Thanks Todd. This is a nice piece to have in place for the upgrades.

PR #268, opened 2022-02-01
Append LIMS linelist to ID3C linelist

davereinhart:
Append LIMS linelist from S3 to ID3C-generated linelist. If the file is empty or not found, the job will proceed without the LIMS linelist data.

This PR includes also includes refactoring of the scripts to be more consistent with others in backoffice. The `transform` script will now only do a transformation (outputting the resulting CSV to `stdout`) and the `upload` task has been moved to a new script.  `generate` has been wrapped in a function with parameter parsing and validation.

A possible future improvement would be to move the REDCap data retrieval from `transform` into its own script, saving the results in `.ndjson` format and then passing that file to `transform`.


davereinhart:
[Append DoH Linelist to include AIRS Data](https://trello.com/c/bxx5flJa/1307-append-doh-linelist-to-include-airs-data)

